PR #59, opened 2025-12-18
Add GitHub Action to protect released schema versions

tdjager:
## Summary
- Adds a GitHub Action that prevents modifications or deletions of released schema versions (`schema/vX.X.X/`)
- Only runs on PRs to main - allows free iteration in feature branches
- Compares PR changes against schema versions that already exist on main
- Adds `schema/latest/` directory that tracks the current schema version (not protected)

## Why GitHub Action instead of pre-commit hook?
A pre-commit hook runs on every commit, which would block iterating on new schema versions during development. The protection should only apply when merging to main - that's when a schema version becomes "released" and immutable.

## Test plan
- [x] CI workflow runs and passes on this PR
- [x] `schema/latest/` can be updated (demonstrated in commit history)
- [x] Adding new schema version (v0.4.0) passes
- [x] Removing schema version not on main (v0.4.0) passes
- [x] Modifying schema version on main (v0.3.0) fails CI
- [x] Restoring schema to original state passes CI (compares net diff, not history)

### DIFF
--- a/schema/latest/schema.json
+++ b/schema/latest/schema.json
@@ -22,8 +22,8 @@
     },
     "valid_from": {
       "type": "string",
-      "pattern": "^[0-9]{4}-[0-9]{2}-[0-9]{2}$",
-      "description": "Date when the law/regulation becomes effective (inwerkingtredingsdatum)"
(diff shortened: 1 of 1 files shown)


anneschuth:
@tdjager do we want to merge this?

PR #176, opened 2026-02-26
Add release workflow for engine binary

anneschuth:
## Summary

- Adds a GitHub Actions workflow (`release-engine.yml`) that cross-compiles the `evaluate` binary for 3 targets when an `engine-v*` tag is pushed
- Adds the `evaluate` CLI binary for law evaluation via stdin/stdout JSON
- Adds an install script (`script/install-engine.sh`) for easy downloading

### Targets
| Asset | Platform |
|-------|----------|
| `evaluate-linux-amd64` | Linux x86_64 |
| `evaluate-macos-amd64` | macOS Intel |
| `evaluate-macos-arm64` | macOS Apple Silicon |

### How to release
```bash
git tag engine-v0.1.0
git push origin engine-v0.1.0
```

### How to install
```bash
# Latest
./script/install-engine.sh

# Specific version
./script/install-engine.sh engine-v0.1.0

# Custom directory
INSTALL_DIR=/usr/local/bin ./script/install-engine.sh
```

### Why
WIAT (and other consumers) needs the `evaluate` binary to run law simulations. Currently it relies on a sibling repo checkout with `cargo build --release`. This enables proper versioned releases so consumers can pin and download prebuilt binaries.

## Test plan
- [ ] Merge, then push `engine-v0.1.0` tag
- [ ] Verify GitHub release is created with 3 binary assets
- [ ] Download each binary on its target platform and run: `echo '{}' | ./evaluate-linux-amd64`
- [ ] Test install script: `./script/install-engine.sh engine-v0.1.0`

### DIFF
--- a/.github/workflows/release-engine.yml
+++ b/.github/workflows/release-engine.yml
@@ -50,14 +50,14 @@ jobs:
             ~/.cargo/registry
             ~/.cargo/git
             packages/target
-          key: ${{ runner.os }}-${{ matrix.target }}-cargo-release-${{ hashFiles('packages/**/Cargo.toml') }}
+          key: ${{ runner.os }}-${{ matrix.target }}-cargo-release-${{ hashFiles('packages/**/Cargo.toml', 'packages/Cargo.lock') }}
           restore-keys: |
             ${{ runner.os }}-${{ matrix.target }}-cargo-release-
 
       - name: Build release binary
         run: |
           cd packages/engine
-          cargo build --release --bin evaluate --target ${{ matrix.target }}
+          cargo build --release --locked --bin evaluate --target ${{ matrix.target }}
 
       - name: Prepare artifact
         run: |


tdjager:
~~Small note on the cache key fix: the suggestion was applied with `packages/Cargo.lock`, but the diff in this PR modifies the **root** `Cargo.lock`. Since `cargo build` in `packages/engine` uses the workspace root lockfile, the cache key should reference `Cargo.lock` (root) instead of `packages/Cargo.lock` to properly invalidate on dependency changes.~~

**Correction**: `packages/Cargo.lock` is correct. Since the workflow does `cd packages/engine` before `cargo build`, Cargo resolves the workspace at `packages/Cargo.toml` (which has its own `[workspace]` with `members = ["engine", "harvester"]`), so `packages/Cargo.lock` is indeed the lockfile used. Apologies for the noise.

Side note: the root `Cargo.lock` changes in this PR are from a separate root workspace that also claims `packages/engine` — having two workspaces that both include the same crate is unusual and could lead to lockfile drift.

tdjager (approved):
LGTM — both review comments have been addressed. The `--locked` flag and cache key fix look good.

PR #315, opened 2026-03-22
feat(frontend): add central corpus as GitHub source in editor

tdjager:
## Summary
- Add `MinBZK/regelrecht-corpus` (development branch) as a GitHub source in `corpus-registry.yaml`
- Extend `copy-laws.js` to fetch GitHub sources via tarball download at build time
- Mount `GITHUB_TOKEN` in Dockerfile build step so the script can authenticate with GitHub API

Local laws (priority 1) override central laws (priority 2) for overlapping `$id`s.

## Test plan
- [x] `node frontend/scripts/copy-laws.js` shows both sources and outputs combined index.json (5730 laws)
- [ ] `npm run build` succeeds with both sources
- [ ] Library page shows laws from both local and central corpus
- [ ] Local laws with same $id as central win (priority resolution)

### DIFF
--- a/frontend/src/LibraryApp.vue
+++ b/frontend/src/LibraryApp.vue
@@ -133,8 +133,9 @@ async function loadIndex() {
       favorites.value = new Set(favIds);
     }
 
-    // Fetch GitHub sources at runtime.
+    // Fetch GitHub sources at runtime (sorted by priority, lower wins).
     let githubIndex = [];
+    const seenIds = new Set(localIds);
     if (registryRes.ok) {
       const registryText = await registryRes.text();
       const registry = yaml.load(registryText);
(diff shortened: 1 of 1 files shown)


tdjager (commented):
LGTM - two minor suggestions.

Previous findings have been addressed. Two new observations below (inline).

frontend/scripts/copy-laws.js
```diff
@@ -79,42 +81,89 @@ function extractMeta(content) {
...
+  const res = await fetch(tarballUrl, { headers, redirect: 'follow' });
+  if (!res.ok) {
+    throw new Error(`GitHub tarball download failed for ${owner}/${repo}@${branch}: ${res.status} ${res.statusText}`);
+  }
+
+  const tmpDir = mkdtempSync(join(tmpdir(), 'corpus-'));
+  try {
+    const tarPath = join(tmpDir, 'repo.tar.gz');
+    writeFileSync(tarPath, Buffer.from(await res.arrayBuffer()));
+    execSync(`tar -xzf ${tarPath} -C ${tmpDir}`, { stdio: 'ignore' });
```
tdjager: **Minor** — `stdio: 'ignore'` on `execSync` swallows `tar` stderr

If the downloaded tarball is corrupt or truncated, `tar` will exit non-zero and `execSync` will throw — but the actual `tar` error message (e.g., "unexpected EOF", "not in gzip format") is silently discarded because `stdio: 'ignore'`. The thrown error only shows "Command failed: tar -xzf ..." with no diagnostic detail.

Consider using `stdio: 'pipe'` so the error message is preserved:
```js
execSync(`tar -xzf ${tarPath} -C ${tmpDir}`, { stdio: 'pipe' });
```

PR #434, opened 2026-04-03
docs: rfc-014 engine conformance test suite

ehotting:
## Summary

- Language-agnostic JSON conformance test suite per schema version
- Conformance levels: Core, Cross-law, IoC, Temporal, Advanced
- Test format allows single and multi-regulation scenarios, error assertions, and optional trace assertions
- Provenance conformance tests verify Execution Receipt fields (RFC-013)
- Initial tests derived from existing BDD scenarios, then extended with edge cases

Depends on: RFC-013 (execution provenance, Execution Receipt format)

## Test plan

- [ ] Review RFC content with team
- [ ] Verify operation coverage matches current schema v0.5.1
- [ ] Verify conformance levels align with RFC-009 multi-org requirements

### DIFF
--- a/packages/engine/src/types.rs
+++ b/packages/engine/src/types.rs
@@ -588,13 +588,13 @@ impl Operation {
         Operation::NotIn,
     ];
 
-    /// All variants of the enum. The exhaustive match in `name()` already
-    /// guarantees compile-time coverage. This array is the single source
-    /// of truth for variant enumeration — its length is the count.
+    /// All variants of the enum. This is a manually maintained list;
+    /// forgetting to add a new variant here compiles fine, but the
+    /// `operation_lists_are_exhaustive` test catches it by cross-checking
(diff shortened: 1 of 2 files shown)


packages/engine/src/types.rs
```diff
@@ -550,6 +550,44 @@ pub enum Operation {
...
+        Operation::Min,
+        Operation::And,
+        Operation::Or,
+        Operation::Not,
+        Operation::If,
+        Operation::In,
+        Operation::List,
+        Operation::Age,
+        Operation::DateAdd,
+        Operation::Date,
```
ehotting: Fixed in becfd0e. Replaced the hardcoded count with `Operation::variant_count()` which uses an exhaustive match internally — adding a new enum variant without updating the match is now a compile error. The unit test then asserts `SCHEMA_OPERATIONS.len() + COMPAT_ALIASES.len() == variant_count()`.

PR #546, opened 2026-04-13
fix(editor): library panel layout and scenario result activation

tdjager:
## Summary
- Fix library article detail panel: add `background="default"` to the sticky-header page so the header is opaque and content doesn't bleed through on scroll
- Add rounded top corners with tinted background to the Machine and YAML panels in the library detail view
- When expanding a scenario accordion in the editor, immediately populate the result panel with that scenario's execution data
- Automatically switch the right pane to "Resultaat" view when scenario execution data arrives

## Test plan
- [ ] Open library, select a law and article, verify content is not hidden behind header/tab bar
- [ ] Switch between Tekst/Machine/YAML tabs and verify Machine and YAML have rounded top corners
- [ ] Open editor, expand a scenario accordion, verify the result panel activates and shows that scenario's result

### DIFF
--- a/frontend/src/components/ScenarioForm.vue
+++ b/frontend/src/components/ScenarioForm.vue
@@ -202,6 +202,14 @@ function matchStatus(outputName, actualValue) {
 }
 
 const hasExpectations = computed(() => Object.keys(expectations.value).length > 0);
+
+const overallStatus = computed(() => {
+  if (!result.value) return null;
+  for (const name of Object.keys(expectations.value)) {
+    if (matchStatus(name, result.value.outputs?.[name]) === 'failed') return 'failed';
+  }
+  return 'passed';
+});
 </script>
 
 <template>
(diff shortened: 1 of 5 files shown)


frontend/src/components/ScenarioForm.vue
```diff
@@ -202,31 +202,43 @@ function matchStatus(outputName, actualValue) {
...
+          <span v-if="articleMap?.outputToArticle?.get(name)" class="sf-article-tag">Art. {{ articleMap.outputToArticle.get(name) }}</span>
+          <span class="sf-expectation-detail">
+            <span>{{ formatValue(normalizeForCompare(exp)) }}</span>
+            <template v-if="result && result.outputs">
+              <span class="sf-expectation-arrow">&rarr;</span>
+              <span>{{ formatOutputValue(result.outputs[name], name) }}</span>
+            </template>
+          </span>
+        </div>
+      </div>
```
tdjager, line 241: Intentional — the user requested a single continuous block with one overall left-border color. Per-item feedback is available in the result pane (ExecutionTraceView) which auto-activates when the accordion opens.

PR #517, opened 2026-04-08
feat(editor): request harvest for missing law dependencies

tdjager:
## Summary

- When the editor loads a law and discovers cross-law references not in the corpus, it automatically requests harvesting with high priority (80)
- Adds `slug` column to `law_entries` table for reverse lookup (slug → BWB ID)
- New `POST /api/corpus/request-harvest` endpoint on editor-api (requires optional `DATABASE_URL`)
- Frontend `useDependencies.js` now reports missing deps and shows harvest feedback in progress message

## How it works

1. Editor loads law → `useDependencies` discovers referenced laws
2. Missing deps are collected after loading phase
3. Frontend calls `POST /api/corpus/request-harvest` with missing slugs
4. Editor-api looks up BWB ID by slug in `law_entries`, creates high-priority harvest job
5. User sees progress: "5/7 wetten geladen — 2 ontbrekende wetten aangevraagd"

## Configuration

- `DATABASE_URL` on editor-api enables harvest requests (optional, graceful degradation when not set)
- Pipeline feature flag on editor-api (`pipeline` feature, enabled by default)

## Test plan

- [x] `cargo check --workspace` passes
- [x] `just format` passes
- [x] `just lint` passes
- [x] `just test` passes
- [x] `just pipeline-test` passes
- [ ] Manual test: open editor with law that has missing deps, verify harvest request is sent
- [ ] Verify pipeline integration test with slug column migration

PR #552, opened 2026-04-15
feat: consolidate harvest into pipeline-api service

tdjager:
## Summary
- Moves BWB search, harvest request, and status tracking out of editor-api into a dedicated `pipeline-api` binary in the pipeline package
- Editor-api proxies `/api/harvest/*` requests to pipeline-api via `PIPELINE_API_URL`, staying stateless
- Frontend uses `useBwbSearch` and `useBwbHarvest` composables instead of ~150 lines of inline logic in LibraryApp.vue
- Replaces raw SQL from PR #517 with existing `job_queue` and `law_status` library functions

## Test plan
- [ ] `cargo check` passes for pipeline, editor-api, and admin packages
- [ ] `cargo clippy` clean for all packages
- [ ] Pipeline unit tests pass (`just pipeline-test`)
- [ ] Start pipeline-api locally, verify `/harvest/search`, `/harvest`, `/harvest/status` endpoints
- [ ] Start editor-api with `PIPELINE_API_URL` set, verify proxy works end-to-end
- [ ] BWB search → harvest → poll → corpus reload flow works in the frontend


tdjager:
Addressed the fix-worthy findings in 9ef94b3 (round 4):
- `update_status_unless_any` now binds `text[]` (via `strum::Display`) so it does not need `PgHasArrayType` on `LawStatusValue`; added integration test covering the protected set.
- Harvest downgrade guard extended to `[Harvesting, Enriching, Harvested, Enriched]` so re-harvesting a finished law does not bounce it back to `Queued`.
- `useBwbHarvest`: `startPolling`/`stopPolling` removed from the public API.
- `MIN_QUERY_LENGTH` exported from `useBwbSearch` and reused in the template.

Declining the two remaining suggestions with reasoning:
- **Hyphen assumption in PR description**: the in-code doc comment above `discover_pipeline_api_url_from_k8s` already carries the assumption and the escape hatch (`PIPELINE_API_URL`). The PR description is about scope; documenting a ZAD-naming convention belongs in the module docstring (where it lives now), not the PR body.
- **Slug exposure on unauthenticated `/api/harvest/status`**: BWB IDs, law slugs, and pipeline status are all public information (slugs map 1:1 to the public URLs on lawmaking.regelrecht.rijks.app and wetten.overheid.nl). Gating this endpoint behind auth would force unauthenticated UI to stop showing harvest progress, which breaks the anonymous-browsing UX. Keeping it public is deliberate.

PR #568, opened 2026-04-17
feat(admin): add fake SSO for PR deployments

tdjager:
## Summary

- Adds `/auth/fake-login` endpoint that creates an authenticated session without real OIDC
- Frontend shows a login choice screen when fake SSO is available (real SSO vs test login)
- Configurable test user via query params (`?name=...&email=...`)
- Production guard: refuses to start if `FAKE_SSO=true` and `DEPLOYMENT_NAME=regelrecht`

## Motivation

In PR deployments the real Keycloak SSO requires 2FA and a government account. This makes it impossible for automated Claude agents to test authenticated endpoints, and prevents testing with different user profiles.

## Configuration

Set `FAKE_SSO=true` as environment variable in PR deployments. No other changes needed.

## Test plan

- [ ] `just build-check` passes
- [ ] `just format && just lint` clean
- [ ] Unit tests pass (5 config tests + 4 auth tests)
- [ ] Manual test: visit admin panel with `FAKE_SSO=true` → login choice screen appears
- [ ] Manual test: click "Test Login" → authenticated as fake user
- [ ] Manual test: real SSO button still works alongside fake login
- [ ] Verify production guard: `FAKE_SSO=true DEPLOYMENT_NAME=regelrecht` → app refuses to start

### DIFF
--- a/.claude/skills/law-download/download_law.py
+++ b/.claude/skills/law-download/download_law.py
@@ -73,6 +73,7 @@ def parse_wti_metadata(wti_tree):
         "GRONDWET", "WET", "AMVB", "KONINKLIJK_BESLUIT", "MINISTERIELE_REGELING",
         "BELEIDSREGEL", "EU_VERORDENING", "EU_RICHTLIJN", "VERDRAG",
         "UITVOERINGSBELEID", "GEMEENTELIJKE_VERORDENING", "PROVINCIALE_VERORDENING",
+        "WATERSCHAPS_VERORDENING",
     }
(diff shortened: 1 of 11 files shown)


tdjager:
Both review findings addressed:

1. **Security (allowlist)**: Changed from blocklist (`deployment == "regelrecht"`) to allowlist — test SSO now only works when `DEPLOYMENT_NAME` matches `pr<digits>`. Any other deployment name (staging, unset, etc.) is rejected. Added 2 tests for this.

2. **Code quality (starts_with)**: Already fixed in previous commit — `is_preview` now validates the full `pr<N>` pattern.

PR #579, opened 2026-04-23
feat(editor): add law dependency graph view (PR1 of 3)

tdjager:
## Summary

- First of three PRs porting Eelco + Anne's `demo/graph/` Svelte app (branch `feature/demo-leenstelsel-tegemoetkoming`) into the editor, so the wettengraaf + scenario-replay becomes part of the editor instead of a standalone demo.
- This PR adds the **static law dependency graph** as a new right-pane option behind feature flag `panel.law_graph` (off by default). No trace stepping yet — that lands in PR2.
- Clicking a law highlights its inbound edges red and outbound edges green; close-button on a law node hides it from the graph.

## Context

See the approved plan discussion on integrating the trace-stepping graph. Full rollout:

| PR | Scope |
|----|-------|
| **PR1 (this)** | Graph rendering + click highlighting, behind `panel.law_graph` flag |
| PR2 | Port `traceEdges.js` + `useTraceStepping` composable + step list + highlights |
| PR3 | Polish: filters, assertions bar, boom-view, performance |

## What's in

- `@vue-flow/core` + `@vue-flow/controls` + `@vue-flow/background` + `@vue-flow/minimap` deps
- `composables/useLawGraph.js` — ported the node/edge builder from `+page.svelte:131-735` (parseLaw, transitive dep walk, two-pass node/edge construction, topological layered layout). Uses `extractRegulationRefs` from `useDependencies.js` rather than re-implementing it. Edge-ID formats (`impl:`, `ovr:`, `hook:`, cross-law input arrows) are byte-identical to the demo so PR2's `traceEdges` matcher works without adaptation.
- `components/LawGraphView.vue` — `<VueFlow>` wrapper + controls/background/minimap, click-to-highlight, hide-law button.
- `components/graph/LawNode.vue` + `LeafNode.vue` — Vue ports of the demo Svelte node components.
- `components/graph/graph-styles.css` — ported demo `:global()` styles; Tailwind `@apply` converted to vanilla CSS with the editor's color tokens. Must be global (Vue Flow renders edges outside the component's scoped-CSS boundary). Trace-stepping CSS deferred to PR2.
- `EditorApp.vue` — adds a 3rd segmented-control option "Graaf" next to "Resultaat" / "Machine" in the right-pane toggle. Right-pane sync falls back cleanly when options are disabled.
- `panel.law_graph` feature flag (off by default) + entry in the Settings sheet.

## What's deliberately NOT in

- `demoGherkin.ts` — editor already has `@cucumber/gherkin`-based parser
- `wasmEngine.ts` — editor already has `useEngine.js` singleton
- Focus-URL-param (`?focus=law_id`) — editor's "currently open law" is the focus
- The bekendmaking-tweede-run logic — leenstelsel-specific, not a generic feature
- `traceEdges.ts` + step controls — PR2

## Test plan

- [ ] Enable "Wettengraaf" in the Instellingen sheet
- [ ] Open a law with dependencies (e.g. `wet_op_de_zorgtoeslag`) in the editor
- [ ] Switch right pane to "Graaf"
- [ ] Verify graph renders: law card(s) with nested Parameters / Input / Output / Delegeert / Implementeert groups
- [ ] Click a law → verify inbound edges turn red, outbound green; click again to clear
- [ ] Click the × on a law → verify its nodes and connected edges disappear
- [ ] Switch back to "Resultaat" → verify trace view still works
- [ ] ZAD preview deploy: verify no deploy timeout (it's a pure frontend change — no migrations, no backend changes)

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -32,6 +32,8 @@
     },
     "node_modules/@babel/generator": {
       "version": "7.29.1",
+      "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.29.1.tgz",
+      "integrity": "sha512-qsaF+9Qcm2Qv8SRIMMscAvG4O3lJ0F1GuMo5HR/Bp02LopNgnZBC/EkbevHFeGs4ls/oPz9v+Bsmzbkbe+0dUw==",
       "dev": true,
       "license": "MIT",
       "dependencies": {
@@ -47,6 +49,8 @@
     },
     "node_modules/@babel/helper-string-parser": {
       "version": "7.27.1",
+      "resolved": "https://registry.npmjs.org/@babel/helper-string-parser/-/helper-string-parser-7.27.1.tgz",
+      "integrity": "sha512-qMlSxKbpRlAridDExk92nSobyDdpPijUq2DW6oDnUqd0iOGxmQjyqhMIihI9+zv4LPyZdRje2cavWPbCbWm3eA==",
       "license": "MIT",
       "engines": {
         "node": ">=6.9.0"
@@ -54,6 +58,8 @@
     },
     "node_modules/@babel/helper-validator-identifier": {
       "version": "7.28.5",
+      "resolved": "https://registry.npmjs.org/@babel/helper-validator-identifier/-/helper-validator-identifier-7.28.5.tgz",
+      "integrity": "sha512-qSs4ifwzKJSV39ucNjsvc6WVHs6b7S03sOh2OcHF9UHfVPqWWALUsNUVzhSBiItjRZoLHx7nIarVjqKVusUZ1Q==",
       "license": "MIT",
       "engines": {
         "node": ">=6.9.0"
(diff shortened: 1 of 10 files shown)


tdjager:
Re: the two findings I'm not addressing in this PR — context for future reference:

**Global CSS side-effect on import** (`LawGraphView.vue` graph-styles.css import): every rule in that file is anchored on `.vue-flow` or `.vue-flow__*` selectors. When `panel.law_graph` is off, no `.vue-flow` element exists in the DOM, so the rules have zero observable effect. A lazy import would force async rendering around VueFlow mount or accept a FOUC the first time the flag is enabled. Trading 2 KB of inert CSS for sync simplicity is the right call for PR1; if bundle size becomes a concern in PR3 this moves to a dynamic `import()`.

**`height: calc(100vh - 180px)` magic number**: the comment above the rule breaks down where the 180px comes from (48+56+48+28 of chrome). Switching to `height: 100%; flex: 1` inside `nldd-page` requires verifying the outer `nldd-page` actually propagates a definite height to its body slot across the editor layout — that's a storybook-component contract test I'd rather not litigate mid-rebase. The documented magic number is stable today and will loudly fail (blank pane) rather than silently if chrome is added, so the maintenance burden is low.

PR #589, opened 2026-04-24
feat(editor): markdown-backed editable Tekst panel

tdjager:
## Summary

- Replaces the read-only left `Tekst` pane with a TipTap-based WYSIWYG editor that round-trips `articles[].text` as markdown.
- Toolbar uses `@minbzk/storybook` components: `nldd-toolbar` / `nldd-icon-button` / `nldd-dropdown`, with Bold, Italic, bullet list, and numbered list buttons.
- Reuses the machine-panel save infrastructure: both panes now share a single `handleLawSave` that PUTs the full law YAML via the existing `useLaw.saveLaw` flow. A dirty-state footer `Opslaan` button appears in whichever pane has pending edits.

## Behaviour notes

- **First-save YAML normalization.** Articles whose `text` is still plain-numbered prose (`"1. …\n\n2. …"`) will be normalized to canonical ordered-list markdown on first save. The engine ignores `text`, so this is functionally lossless, but the YAML diff will look substantial for articles edited for the first time.
- **Active-state styling is local.** `nldd-icon-button` has no `pressed` attribute today, so each format button is wrapped in a span that adds a subtle tint via scoped CSS. Upstreaming a proper pressed state to `MinBZK/storybook` is a follow-up.
- **`nldd-button-bar-divider` requires `nldd-button-bar`.** Outside that context it renders 0×0, so the separator between the inline-mark and list toggles is a local 1×20px styled `<span>`.
- **`ArticleText.vue` stays.** Still consumed read-only by `LibraryApp.vue` for the library detail view.

## Critical files

- `frontend/src/components/ArticleTextEditor.vue` — new
- `frontend/src/EditorApp.vue` — adds `editedText`, extends the `currentLawYaml` splice + dirty computed, factors `handleLawSave`, swaps the left-pane contents, adds the text-pane footer save button
- `frontend/package.json` — adds `@tiptap/core`, `@tiptap/vue-3`, `@tiptap/starter-kit`, `tiptap-markdown`

## Test plan

- [x] `npm test` (97 tests passing in frontend vitest suite)
- [x] `npm run build` — clean build, bundle grew ~60 KB gzipped (lazy-loaded `EditorApp` route chunk)
- [x] Manual smoke test via Playwright against vite dev server: bold / italic / bullet list / numbered list all toggle correctly, active-state on the toolbar reflects selection, footer `Opslaan` appears when dirty, state re-seeds cleanly on reload
- [ ] End-to-end save-to-disk verified in a preview environment (same `saveLaw` PUT as machine_readable, which is already production-proven)
- [ ] Library detail view still renders `ArticleText.vue` unchanged

## Follow-ups (out of scope)

- Upstream an `nldd-rich-text-editor` (or at least a toggled-state `nldd-icon-button`) to `MinBZK/storybook` so the local scoped active-state and divider hacks can go away.
- One-shot script to migrate existing corpus `articles[].text` to canonical markdown so the first-save normalization isn't distributed across future edits.

### DIFF
--- a/docs/rfcs/rfc-010.md
+++ b/docs/rfcs/rfc-010.md
@@ -263,15 +263,64 @@ corpus-registry.yaml
 
 #### Write path (editor)
 
-The editor can write directly to a GitHub repo via the GitHub Contents API, without a backend intermediary.
+> **Amended after implementation (Phase 6).** The original sketch had the
+> editor speak the GitHub Contents API directly from the browser. We
+> reversed that decision and route writes through `editor-api`. The
+> tradeoffs below explain why.
+
+`editor-api` mediates every write. The frontend does not know which
+GitHub repo backs a given law — it just calls
+`PUT /api/corpus/laws/{law_id}` and the backend resolves the source
+from the in-memory `SourceMap`, looks up the per-source token from the
+existing auth config (env var or `corpus-auth.yaml`), and pushes /
+opens the PR upstream.
 
 Flow:
-1. User edits YAML in the editor
-2. Preview and validation against the schema
-3. Commit via `PUT /repos/{owner}/{repo}/contents/{path}` using the GitHub token from the browser
-4. Branch management: the editor works on a feature branch and can create a PR via the GitHub API (`POST /repos/{owner}/{repo}/pulls`)
+1. User edits YAML in the editor.
+2. Preview and validation against the schema.
+3. The frontend sends `PUT /api/corpus/laws/{law_id}` (or
+   `/scenarios/{filename}` for scenario edits) with an
+   `X-Editor-Session: <uuid>` header. The UUID is minted in
+   `sessionStorage` on first use and is stable for the browser tab.
+4. `editor-api` looks up the law's source. For a GitHub source it
+   resolves (or lazily creates) a `SessionGitBackend` keyed on
+   `(session_id, source_id)`. For a local source it falls through to
+   the existing local-write path so the dev loop is unchanged.
+5. The session backend writes the file into its working clone, calls
+   `commit_and_push_to_branch` (always rebases on the source's
+   configured branch, then fast-forward pushes the session branch
+   `editor/session-<uuid>`), and finally `ensure_pr` against the source
+   repo's Pulls API. Subsequent saves in the same session add commits
+   to the same branch and reuse the same PR.
+6. The save response is JSON `{ pr: { url, number, branch } | null }`.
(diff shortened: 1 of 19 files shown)


tdjager:
Re: review on 79afdf0 — disputing two findings:

**🟠 #2 (Per-session disk clones / SessionRegistry entries never cleaned up):** factual dispute on scope, not on the substance. A correct fix needs a background reaper task, an eviction policy (LRU vs TTL, what counts as "session done"), and likely cross-replica coordination — that's a feature addition, not a line-level fix, and it sits outside the markdown-editor scope of this PR. The inline comment already flags this as a follow-up, and the federated-write-back design assumes a separate cleanup pass. I'll capture it as a tracking issue post-merge rather than expanding this PR.

**🟡 #4 (selectionTick side-effect read + lint suppression):** the current pattern (`selectionTick.value;` to register a reactive dependency) is intentional Vue idiom for "I depend on this ref's identity, not its value" and the suggested refactor (computed Set of active mark names) reshapes the data flow rather than fixing a bug. The lint suppression is a permanent annotation on a stable Vue API, not a workaround. Re-flagging if it actually breaks under a future Vue or eslint-plugin-vue version is fine; refactoring it preemptively when the current code is correct isn't a net win.

Fixes #1 and #3 pushed in fa78cfc.

tdjager:
Re: review on fa78cfc — fixes pushed in f22ba10, plus a third finding I'm disputing:

**🟡 \`pr_client.rs\` — one reqwest::Client per session backend:** disputing as out-of-scope, in line with your own framing ("for the expected number of concurrent sessions this is fine"). Sharing a single Arc<PullRequestClient> across all SessionGitBackend instances requires plumbing it through AppState into SessionRegistry::resolve_session_backend and into the SessionGitBackend constructor — a non-trivial signature change with no measurable benefit at current load. Worth doing alongside the session-cleanup follow-up the registry doc-comment already flags; tracking together as a single "session lifecycle hardening" issue makes more sense than splitting it from the cleanup work.

**Fixed:**
- \`ArticleTextEditor.vue\` selectionTick side-effect: replaced with an \`activeFormats\` computed map; eslint-disable removed; adding a new toolbar mark name is now one line.
- \`EditorApp.vue\` \`currentLawYaml\` cosmetic re-dump: added a pristine-article short-circuit returning \`rawYaml.value\` verbatim when neither pane has diverged from the parsed base. \`handleActionSave\` no longer produces reformat-only commits on pristine text-only articles.

PR #591, opened 2026-04-27
feat(editor): add per-user settings with theme toggle

tdjager:
## Summary
- Persists per-user editor settings (keyed by OIDC `person_sub`) in a new `user_settings` table; first setting is **theme** (light/dark).
- Frontend mirrors the chosen theme onto `<html data-scheme="...">`, which is the attribute `@minbzk/storybook` (NDD) already keys its dark tokens on — light/dark applies across editor + library views without bespoke CSS.
- Theme toggle ("Donkere modus") added as a checkbox menu-item in the existing account-menu (person-circle).
- Fail-soft loading: 401 (auth off), 503 (no DB), and network errors collapse to `prefers-color-scheme` defaults, so a brand-new or anonymous user always lands in a working editor.

## Implementation notes
- Mirrors the favorites pattern: directly-bound sqlx queries in `editor-api`, schema migration in `packages/pipeline/migrations/`, `person_sub` from session via `regelrecht_auth::SESSION_KEY_SUB`.
- Allowed-keys list (`ALLOWED_KEYS = ["theme"]`) and per-key value validation prevent the table from becoming a free-form key/value bag.
- Idempotent upsert (`ON CONFLICT DO UPDATE`) — first write creates the row, no setup needed for a new user.

## Test plan
- [x] `just format`, `just lint`, `just test` — all green; 3 new unit tests for `validate()`.
- [x] Migration 0013 applied cleanly on Postgres 17 from a fresh schema; trigger fires on `UPDATE` and on upsert via `ON CONFLICT DO UPDATE`.
- [ ] End-to-end in dev (Postgres + editor-api + Vite): new user lands with system-default theme, toggle persists across hard-refresh, library view picks up the theme, `PUT /api/user/settings/theme {"value":"purple"}` → 400, anonymous request → 401.
- [ ] Visual check: dark tokens from `@minbzk/storybook` actually flip the editor and library palettes.

### DIFF
--- a/frontend/src/EditorApp.vue
+++ b/frontend/src/EditorApp.vue
@@ -6,6 +6,7 @@ import { useLaw, fetchLaw } from './composables/useLaw.js';
 import { useEngine } from './composables/useEngine.js';
 import { useAuth } from './composables/useAuth.js';
 import { useFeatureFlags } from './composables/useFeatureFlags.js';
+import { useUserSettings } from './composables/useUserSettings.js';
 import ArticleText from './components/ArticleText.vue';
 import ActionSheet from './components/ActionSheet.vue';
 import EditSheet from './components/EditSheet.vue';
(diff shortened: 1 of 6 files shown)


tdjager:
Replaced by #592 to bypass a stuck ZAD-state on the `harvester-admin × pr591` preview deploy. All commits and review history are carried over.

PR #603, opened 2026-05-01
refactor: dedup hot-path patterns across crates and frontends

tdjager:
## Summary

Five-commit DRY pass across the workspace. Each commit is independently reviewable; together they remove ~600 net lines of duplication while preserving behaviour. All checks green: ` just format`, `just lint`, `just test`, `just pipeline-test`.

### What changed

1. **\`refactor: small DRY cleanups across crates\`**
   - \`CorpusRegistry::empty()\` replaces three inline empty-YAML-with-\`unreachable!\` blocks in admin/editor-api.
   - \`[lints] workspace = true\` on every crate replaces nine duplicates of \`[lints.clippy]\`.
   - Engine \`Value\` From-impl pair: extract \`json_number_to_value\` and \`json_object_as_untranslatable\` (the only genuinely-shared blocks; move/clone arms intentionally stay separate).
   - Pipeline: extract \`fail_count_column(JobType)\` shared by \`increment_fail_count\` / \`reset_fail_count\`.

2. **\`build: hoist common deps into [workspace.dependencies]\`**
   - Centralises serde / tokio / tracing / sqlx / axum / chrono and ~25 others.
   - Aligns \`pretty_assertions\` (\`1\` → \`1.4\`), \`tempfile\` (\`3\`/\`3.27\` → \`3\`).
   - Note: \`reqwest\` is intentionally **not** hoisted — admin/auth/editor-api are pinned to 0.12 by openidconnect/oauth2's transitive dep, while corpus/harvester/pipeline track 0.13. Aligning needs an openidconnect bump (out of scope).

3. **\`refactor: consolidate shared infra (telemetry, health server, TestDb)\`**
   - \`regelrecht_shared::telemetry::init_subscriber\` (behind \`telemetry\` feature) replaces five identical \`tracing_subscriber::fmt\` blocks across binaries.
   - \`regelrecht_pipeline::health::spawn_health_server\` replaces the two hand-rolled HTTP/1.1 health endpoints. Settles on enrich_worker's \`HEALTH_PORT\` convention; harvest_worker no longer hardcodes 8000.
   - \`regelrecht_pipeline::test_utils::TestDb\` (behind \`test-utils\` feature) hosts the testcontainers Postgres helper that was byte-identical between admin and pipeline integration tests. Both \`tests/common/mod.rs\` deleted.

4. **\`refactor: dedup hot-path patterns across crates\`**
   - \`regelrecht_corpus::dto\`: shared \`PaginationParams\`, \`SourceSummary\`, \`build_source_summaries\`, \`source_type_tag\`. admin and editor-api stop redeclaring these.
   - admin: route date validation through \`regelrecht_harvester::validate_date\` — fixes a real divergence where admin accepted future dates the harvester would reject.
   - admin: \`db_err(op)\` helper replaces ~12 copies of \`.map_err(|e| { tracing::error!(error = %e, …); ApiError::Internal(\"internal server error\".into()) })\`.
   - engine: \`int_or_float_result\` shared by ADD/SUBTRACT/MULTIPLY's promotion-and-overflow tail.
   - pipeline: \`claim_job\` collapses to a single SQL gated by \`($1::job_type IS NULL OR job_type = $1)\`.
   - harvester: \`LidHandler\`/\`LiHandler\` parameterised over a skip-tag list; \`Considerans\`/\`Aanhef\` over a separator. \`convert_jci_to_url\` replaces four anchor branches with a (regex, prefix) table.
   - tui: \`views::common::titled_block\` covers the \`Block::default().borders(Borders::ALL).title(Span::styled(…, bold))\` pattern that 11 panel renders inlined.

5. **\`refactor(frontend): tokenise brand colour, drop dead component, dedup helpers\`**
   - \`--color-primary: #154273\` token in editor frontend, replacing eight inline hex uses across DataSourceTable / ScenarioGherkin / ScenarioVisual.
   - Delete unimported \`ScenarioResults.vue\` (200+ lines, also reimplemented helpers from outputFormat.js with subtly different semantics).
   - \`structuredClone\` replaces seven \`JSON.parse(JSON.stringify(...))\` deep clones.
   - \`authedFetch\` helper in admin frontend (composables/useAuth.js) replaces five copy-pasted \`if (response.status === 401) { redirectToLogin(); return; }\` blocks.

## Behavioural notes

- \`harvest_worker\` now respects \`HEALTH_PORT\` (was hardcoded 8000). Existing deployments without that env var keep the same effective port; deployments that set it now get honoured.
- admin's date validation gets stricter (rejects future dates) — this is a bug fix matching the harvester's contract.

## Test plan

- [x] \`just format\` — clean
- [x] \`just lint\` — clean
- [x] \`just build-check --all-features\` — clean
- [x] \`just test\` — all suites pass (325 engine, 12 BDD, 31 pipeline, 4 trace, 2 docs, 3 untranslatable)
- [x] \`just pipeline-test\` — all 31 pipeline unit tests pass
- [ ] CI green
- [ ] Frontend builds (no node_modules in worktree; verified statically — \`grep -r ScenarioResults frontend/\` returns no imports, structuredClone is browser-stable, authedFetch is a clean drop-in)


tdjager:
Re: claim_job plan-sharing — agreed it's a real tradeoff, but the dedup was the explicit goal of the change (the two-branch SQL was the duplication this PR was hired to remove). The reviewer notes "for the current job volumes this won't matter", and the queue is bounded by harvest/enrich job rates that are orders of magnitude below where a generic-plan vs branch-plan difference would matter on a partial index. Reverting would re-introduce the duplication for a hypothetical future scaling concern. If `claim_job` ever does become hot, the right fix is benchmark-driven and probably involves a different structure (per-type queues, listen/notify) rather than restoring the duplicate SQL.

tdjager:
Both findings noted, disputing both:

**Finding 1 (claim_job query plan):** Hypothetical/conditional. The reviewer's own assessment is "negligible at current queue sizes" and the mitigation is gated on a `(status, job_type, priority, created_at)` index being "ever added" — which doesn't currently exist, so there is no index for the planner to skip. The suggested cast removal (`$1 IS NULL OR job_type = $1`) would fail because Postgres requires explicit type information on NULL parameters of custom enum types under sqlx's extended protocol. Reverting to two queries would undo the deliberate dedup that is the whole point of this PR. Will revisit if a real `(status, job_type, …)` index is introduced and benchmarks show plan regression.

**Finding 2 (testcontainers in `[dependencies]`):** Reviewer-acknowledged ("this is acceptable as-is"). The optional + feature-gate pattern is the canonical workaround for Cargo's missing workspace-level dev-dependencies — there is no other way to share `TestDb` across crates. `test-utils` is off by default, so `cargo build` (without `--all-features`) does not pull testcontainers; SBOM impact is contained to dev-only build paths.

tdjager:
Re: 'Admin POST /api/harvest-jobs is now stricter (future dates rejected)' — this behavior change is intentional and advertised in the PR description / commit messages: the admin endpoint now delegates to `regelrecht_harvester::validate_date` so the validation is consistent across the codebase. No code change requested by the reviewer; flagging this as informational only.

For the other two findings:
- 🟠 `HEALTH_PORT` silent swallow → fixed: `health_port()` now logs a `tracing::warn!` when the env var is set but unparseable.
- 🟡 `sed`-based member trimming → fixed: extended the existing comments above each `sed` block in the 5 Dockerfiles to explicitly call out the dependency on `packages/Cargo.toml` listing each member on its own line.

tdjager:
Round 3 triage:

- 🟡 Sed lacks maintenance contract → fixed: added one-line note to each Dockerfile sed block reminding contributors to update the exclusion list when adding a workspace member this image doesn't build.
- 🟡 `claim_job` unified SQL plan → disputing. The reviewer's own framing is 'fine at current volumes, worth documenting *if* claim_job ever shows up in a slow-query log.' The action item is conditional on a future signal that hasn't materialised, and adding speculative perf commentary to the SQL would just be noise that needs revisiting if the perf profile actually changes. The unification was deliberate and the trade-off is well-understood.

Also acknowledging the reviewer's self-correction on the prior round's findings about `health_port()` and `init_subscriber` — those were already correctly fixed in round 2 and the prior reviews mis-read the diff.

PR #626, opened 2026-05-13
feat(admin): layered RBAC with per-app reader/writer/admin roles

tdjager:
## Summary

- Introduces a four-layer role model (`platform-admin > app-admin > app-writer > app-reader`) gated per route, replacing the single `allowed-user` login check.
- Hierarchy lives in Keycloak as **composite roles**: each route asks for exactly one role, and higher roles automatically carry lower ones in the JWT — no `role-A OR role-B` logic in code.
- Two apps today: `editor` (editor-api + frontend) and `harvester` (admin dashboard). A `platform-admin` role spans both.

## Role mapping

**editor-api** (deployed with `OIDC_REQUIRED_ROLE=editor-reader`)

| Routes | Required role |
|---|---|
| `GET /api/sources`, `GET /api/corpus/laws*`, `GET /api/feature-flags`, `GET /api/harvest/status`, `/health` | public |
| `GET /api/favorites`, `GET /api/user/settings`, `GET /api/harvest/search` | `editor-reader` |
| `PUT/DELETE /api/favorites/*`, `PUT /api/user/settings/*`, `PUT /api/corpus/laws/{id}`, `PUT/DELETE /api/corpus/laws/{id}/scenarios/{file}`, `POST /api/harvest`, `POST /api/harvest/batch` | `editor-writer` |
| `POST /api/corpus/reload`, `PUT /api/feature-flags/{key}` | `editor-admin` |

**admin / harvester** (deployed with `OIDC_REQUIRED_ROLE=harvester-reader`)

| Routes | Required role |
|---|---|
| `GET /api/jobs*`, `GET /api/law_entries`, `GET /api/sources`, `GET /api/corpus/laws`, `GET /api/info` | `harvester-reader` |
| `POST /api/harvest-jobs`, `POST /api/enrich-jobs` | `harvester-writer` |
| `POST /api/law_entries/{id}/reset-exhausted`, `POST /api/sources/{id}/sync`, `DELETE /api/jobs` | `harvester-admin` |

The bearer API-key bypass on GET/DELETE in the admin app is unchanged; it now functions as a `platform-admin`-equivalent for those methods.

## Specific (orthogonal) rights

Single sensitive functions can have their own realm role (e.g. `editor-publish`), added to the `<app>-admin` composite so admins inherit it implicitly. Pattern documented; no concrete specific rights introduced in this PR.

## Test plan

- [x] Unit tests in `packages/auth` (40) and `packages/admin` (60) green, including new coverage for: reader→writer 403, writer→reader passthrough via composite, app-admin and platform-admin transitivity, specific-right grant/withhold, no-roles-after-auth 403.
- [x] `just format`, `just lint`, `just build-check`, `just validate` clean.
- [ ] Admin integration tests (`handler_tests`) require testcontainers + a reachable Docker host; CI will run them.
- [ ] **Operator pre-merge**: configure the seven realm roles + composite chains + ID-token mapper in Keycloak, assign existing users an appropriate role, and set `OIDC_REQUIRED_ROLE` per component in ZAD before this lands.
- [ ] PR-preview smoke test with accounts at each role tier.

## Migration

See `docs/operations/auth.md` for the lockout-free rollout: stage Keycloak first, deploy with `OIDC_REQUIRED_ROLE=allowed-user` as a transitional value, then flip to the new role per component, then remove `allowed-user`.

## Out of scope

- Frontend role-awareness (hiding buttons the user can't use).
- Further domain scoping (per-ministerie, per-law).
- Specific rights for concrete features — the mechanism lives here, but which features get one is decided per-feature later.


tdjager:
Re: `Box::pin` per request in middleware factories — disputed.

`Box::pin` here is the idiomatic axum middleware pattern for `from_fn_with_state` factories that return a captured argument (the `required_role`/`role` parameter). The factory must return an `impl Fn(...) -> impl Future<...> + Send`, and the inner async block captures the role by move — type-erasing the future via `Box::pin` is the standard way to express this without naming an unnameable future type. Eliminating the allocation would require hand-rolling a concrete `Future` state machine or moving to a hand-written `Layer` + `Service` pair, which is substantial complexity. Reviewer notes this is informational at our request volumes; agreed it is not actionable.

tdjager:
Re: `RequireAuthFuture` / `RequireRoleFuture` aliases duplicated — disputed.

The two type aliases differ in their `Output` type: `RequireAuthFuture` returns `Result<Response, ApiError>` (admin) and `RequireRoleFuture` returns `Result<Response, StatusCode>` (auth). Unifying them would require a generic alias like `type RequireMiddlewareFuture<E> = Pin<Box<dyn Future<Output = Result<Response, E>> + Send>>` — but each call site would still have to specify `<ApiError>` or `<StatusCode>`, providing no real abstraction. Keeping the per-crate aliases keeps usage sites read naturally and isolates concerns to each crate.

PR #632, opened 2026-05-15
feat: introduce traject concept for grouped law edits

tdjager:
## Summary

Introduces the **traject** concept: a named project that groups law edits by one or more people, with its own stable branch on the corpus repo. Replaces the previous per-browser-session `X-Editor-Session` save flow entirely.

A traject has:
- **name + description + scope** (scope is free-form text for now — we'll formalise law links later)
- **status**: `bezig` / `afgerond`
- **owner + members**, each with a role (`beheerder` / `lid`)
- **its own federated corpus config** in the database (mirrors `corpus-registry.yaml`), with exactly one source flagged `is_writable_own = TRUE` at priority 0 — that's where the traject's edits push
- a branch on that writable source, auto-derived from `traject/<slug>-<short-id>` unless an explicit one is supplied

Reads inside a traject see the seeded sources + the traject's own writable source layered on top, so edits made within the traject are visible immediately.

## :rotating_light: Breaking change

**Editor saves now require an active traject.** Without one:
- the editor renders read-only (`canEdit` requires `activeTrajectId !== null`)
- save handlers return `403 Forbidden` with `\"Selecteer eerst een traject om te bewerken\"`

The old `X-Editor-Session` header, the `SessionRegistry`, and the per-(session, source) on-disk clone fan-out are **gone**. There is no fallback save path — picking a traject is the only way to write.

## Endpoints added

| Method | Path | Notes |
|--------|------|-------|
| `GET`    | `/api/trajects`                              | List trajects the caller is a member of |
| `POST`   | `/api/trajects`                              | Create + seed sources |
| `GET`    | `/api/trajects/:id`                          | Detail (members + sources) |
| `PATCH`  | `/api/trajects/:id`                          | Update metadata / status — beheerder-only |
| `DELETE` | `/api/trajects/:id`                          | Hard delete (cascades to members + sources) — beheerder-only |
| `POST`   | `/api/trajects/:id/leave`                    | Self-remove; last beheerder gets 409 |
| `POST`   | `/api/trajects/:id/members`                  | Invite an existing account by email |
| `PATCH`  | `/api/trajects/:id/members/:account_id`      | Change role |
| `DELETE` | `/api/trajects/:id/members/:account_id`      | Remove member |
| `GET`    | `/api/session/active-traject`                | Read active traject id |
| `PUT`    | `/api/session/active-traject`                | Set/clear (verifies membership) |

## Schema

`0014_trajects.sql` adds four tables and three enums:
- \`accounts\` — stable UUID per OIDC subject, upserted on auth (so favorites/settings/trajects can FK by id instead of by the opaque \`sub\`)
- \`trajects\` — name/description/scope/status/created_by
- \`traject_members\` — \`(traject_id, account_id, role)\`
- \`traject_corpus_sources\` — DB mirror of the YAML registry, one row per source, partial unique index enforces \"exactly one writable_own per traject\"

## Frontend

- New \`TrajectMenu.vue\` dropdown in all three responsive toolbars (md/lg/sm): switch traject, clear active, open the create dialog (name/description/scope + writable GitHub source fields).
- New \`useTrajects.js\` composable holds the list + active id as module-singleton state.

## Out of scope (deliberate, follow-up PRs)

- Members-management UI (backend exists; add/remove members is API-only today).
- Working-clone reaper for \`/tmp/regelrecht-editor-corpus/trajects/...\`.
- Per-traject documents in blob/file storage.
- Branch cleanup on \`DELETE /api/trajects/:id\` — we leave the upstream branch behind on purpose; killing it would require a manual call.
- \`/api/me\` to expose the resolved \`AccountRecord\` to the frontend.

## Test plan

- [ ] \`just check\` passes (format + lint + check + validate + unit tests).
- [ ] \`TESTCONTAINERS_HOST_OVERRIDE=host.docker.internal just editor-api-integration-test\` — 17 tests for the traject handlers (CRUD, role guards, last-beheerder guard, cascade delete, session clear on leave). Run unset in CI where 127.0.0.1 works.
- [ ] Manual smoke against a running stack: create a traject, set it active, edit a law, confirm the commit lands on \`traject/<slug>-<short>\` on the writable source. Without an active traject: confirm the editor renders read-only and \`POST /api/corpus/laws/...\` returns 403.


humanintheloop25:
Hi @tdjager — FYI: #653 bouwt een docs-reader bovenop deze PR. Backend staat nu op stub-data, schakelt over naar `traject_members` + `traject_corpus_sources` zodra dit + #626 mergen. Geen haast — we volgen je schema.

PR #661, opened 2026-05-20
feat(editor): invite-by-email for trajects with pending invites

tdjager:
## Summary
- Renames the traject role enum from `beheerder/lid` to `owner/contributor` (PR #632 had it in Dutch).
- Adds a `traject_invites` table so owners can invite **any** email — known or unknown. Migration 0014's "no stub accounts" principle is preserved; OIDC is still the only path that creates `accounts` rows.
- `account_middleware` now promotes any pending invites matching the user's email after the upsert, so a freshly-invited user sees the traject on the next page load (no logout/login required).
- New UI: a **Beheer leden…** entry in the `TrajectMenu` dropdown opens a side-sheet showing active members, pending invites, and an invite form. Owner-only role/remove controls; non-owners see a *Verlaat traject* affordance.

## Background
On `origin/main` after PR #632 the membership flow already exists, but `add_member` rejects any email without an existing account (404). That blocks the typical onboarding flow — "invite a colleague by email." This PR is the smallest cut that fixes that and gets us a real members UI, without disturbing the existing layered RBAC (`editor-writer` app role still gates the API surface).

## Key changes
- `packages/pipeline/migrations/0015_traject_invites.sql` — enum rename + new `traject_invites(traject_id, email, role, invited_by, invited_at)` keyed on `(traject_id, email)`.
- `packages/editor-api/src/trajects.rs` — `add_member` splits on known/unknown email and returns `{status: "active" | "pending", email}`; new `remove_invite` handler; `TrajectDetail.pending_invites` exposed via `GET /api/trajects/:id`; helpers renamed to `require_owner`; email normalisation centralised in `normalize_email`.
- `packages/editor-api/src/accounts.rs` — new `promote_pending_invites` runs in `account_middleware` after the account upsert; ON CONFLICT DO NOTHING + unconditional DELETE keeps the path idempotent. Failures log but don't 500 the request.
- `packages/editor-api/src/main.rs` — register `DELETE /api/trajects/:id/invites/:email` under the same `editor-writer` + `account_middleware` stack.
- `frontend/src/components/TrajectMembersDialog.vue` + `frontend/src/composables/useTrajectMembers.js` — new side-sheet and the composable that owns members/invites state.
- `frontend/src/components/TrajectMenu.vue` — *Beheer leden…* entry, teleports the dialog at the menu's level.

## Authz model (unchanged in shape)
Two-layer: **app role** (`editor-writer` realm role gates the route group) AND **traject role** (`owner` can manage members, `contributor` cannot). Inviting someone requires both. The `accounts.email` UNIQUE invariant is preserved; emails are lowercased on both sides of every comparison.

## Test plan
- [x] `just format` clean
- [x] `just lint` clean (clippy with `-Dwarnings`)
- [x] `just build-check` clean
- [x] `just test` — 191 unit tests pass
- [x] `just editor-api-integration-test` — **23 traject tests pass**, including 7 new ones:
  - `add_member_with_unknown_email_creates_pending_invite`
  - `add_member_normalizes_email_case_for_pending_invites`
  - `pending_invite_is_promoted_to_membership_on_login`
  - `promotion_is_idempotent_when_user_is_already_a_member`
  - `get_returns_pending_invites_alongside_members`
  - `remove_invite_deletes_and_is_idempotent_404`
  - `remove_invite_requires_owner`
- [x] `frontend` build clean; vitest: 191/191 passing
- [ ] Manual against Keycloak dev realm with two test users — verify the invite/promote round-trip end-to-end through the UI
- [ ] Verify ZAD `prN` preview deploy comes up green (migration 0015 applies cleanly)

## Out of scope (deliberate)
- Email notifications for invites (no mail server yet)
- Address book / "users from my org" suggestions — will hook in as a separate composable when organisations land
- Link-sharing or invite shortcodes
- `reviewer` and other future roles — start with `owner/contributor` only

### DIFF
--- a/frontend/src/components/TrajectMembersDialog.vue
+++ b/frontend/src/components/TrajectMembersDialog.vue
@@ -1,0 +1,354 @@
+<script setup>
+import { computed, nextTick, ref, watch } from 'vue';
+import { useTrajectMembers } from '../composables/useTrajectMembers.js';
+import { refreshTrajects } from '../composables/useTrajects.js';
+
+const props = defineProps({
+  /** Whether the sheet is currently open. */
+  modelValue: { type: Boolean, default: false },
+  /** Traject to manage. */
+  trajectId: { type: String, default: null },
+  /** Traject display name, for the sheet header. */
+  trajectName: { type: String, default: '' },
+});
+
+const emit = defineEmits(['update:modelValue']);
+
+const sheetEl = ref(null);
+
+const {
+  members,
+  pendingInvites,
+  callerRole,
+  loading,
+  error: loadError,
+  load,
+  invite,
+  updateRole,
+  removeMember,
+  removeInvite,
+  leaveTraject,
+} = useTrajectMembers();
+
+const isOwner = computed(() => callerRole.value === 'owner');
+
+// Invite form
+const inviteEmail = ref('');
+const inviteRole = ref('contributor');
+const inviteBusy = ref(false);
+const inviteError = ref(null);
+const inviteResult = ref(null);
+
+// Row-level busy + error indicators keyed by member account_id / invite email.
+const rowBusy = ref(new Set());
+const rowError = ref(new Map());
+
+function markBusy(key, busy) {
+  const next = new Set(rowBusy.value);
+  if (busy) next.add(key);
+  else next.delete(key);
+  rowBusy.value = next;
+}
+
+function setRowError(key, message) {
+  const next = new Map(rowError.value);
+  if (message) next.set(key, message);
+  else next.delete(key);
+  rowError.value = next;
+}
+
+watch(
+  () => props.modelValue,
+  async (v) => {
+    await nextTick();
+    if (v) {
+      inviteEmail.value = '';
+      inviteRole.value = 'contributor';
+      inviteError.value = null;
+      inviteResult.value = null;
+      rowBusy.value = new Set();
+      rowError.value = new Map();
+      if (props.trajectId) {
+        await load(props.trajectId);
+      }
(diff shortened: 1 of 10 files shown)


tdjager:
Re: `invited_by` FK defaults to RESTRICT (review 4328515934, finding 3) — disputing.

Migration 0014 takes the same posture for every account reference: `trajects.created_by UUID NOT NULL REFERENCES accounts(id)` and `traject_members.account_id UUID NOT NULL REFERENCES accounts(id)` both omit an `ON DELETE` clause and so default to RESTRICT. Changing only `traject_invites.invited_by` would create cross-table inconsistency where deleting an account is restricted by some referencing tables and not by others — that's harder to reason about than the current uniform "you can't delete an account that's still referenced anywhere".

There's no account-deletion feature today (and accounts are upserted by `person_sub`, never deleted). When one lands it should pick a single cross-table policy (SET NULL for provenance columns, CASCADE for ownership-style columns, or a soft-delete flag) and apply it in one migration. Bundling that decision with this PR would expand scope past invite-by-email; leaving it for the account-deletion PR keeps the schema's deletion policy consistent meanwhile.

tdjager:
Re: `invited_by` ON DELETE RESTRICT (review 4328618914, finding 1) — same disposition as my earlier comment on review 4328515934 finding 3. The implicit RESTRICT is intentional and matches `trajects.created_by` and `traject_members.account_id` in migration 0014, which use the same notation. The right place to switch to explicit RESTRICT + a comment (or to SET NULL) is when account deletion lands, in a single migration that touches all three columns together — not piecemeal here on the new column only.

PR #736, opened 2026-06-03
feat(editor): groepeer bibliotheek-artikellijst per artikel

tdjager:
## What & why

The library's second column (article list) was filled 1:1 from `articles[]`. For deeply split laws this rendered every segment as its own item — e.g. `Artikel 1.e.1°` as a separate menu entry for [`besluit_zorgverzekering`](https://editor.regelrecht.rijks.app/library/besluit_zorgverzekering). The list now groups **per article**: all segments of one article fold into a single item, and selecting it shows them together.

## Grouping key: the url anchor

Every entry has a required `url` with an `#Artikel…` anchor — the real article anchor (= the harvester's `number_parts[0]`). This is more reliable than the dotted number:

- `1.e.1°`, `1.f` → `#Artikel1`
- `2.4.1`, `2.4.1.a.1°` → `#Artikel2.4`
- a genuine dotted article like `3.1` → `#Artikel3.1` (its own group, **not** over-grouped)

No schema, harvester, or corpus changes needed.

## Part 1 — Library (read-only)

- `utils/articleGroups.js` — shared `groupArticles` + `findGroupForArticleNumber` (+ tests)
- `components/ArticleGroup.vue` — stacks the segments, reusing `ArticleText` / `MachineReadable` / `YamlView`; the text reads as one continuous article, `machine_readable` is shown per segment with its segment number as provenance (+ tests)
- `LibraryApp.vue` — column 2 groups, the detail renders the group, and deep-links to a leaf number normalize to the article number

## Part 2 — Editor

Without this, "Edit" landed on the first segment and the underlying parts (`1.a`, `1.b`, `2.4.1.a.1°`) became unreachable in the editor (which has no article list — only tabs + search).

- **Tabs are now per article** (group number) instead of per leaf segment: one tab per article, remembering which segment is active — no tab spam when switching parts.
- An **"Onderdeel" (part) selector** in the document-tab bar (only for split articles) makes every segment reachable and editable.
- Editing/saving still operates on the active segment (`selectedArticleNumber`) — grouping is purely a navigation layer; the splice/PUT logic is unchanged.

Fully seamless stacked text editing (all segments as one continuous editor) was deliberately left out: the editor's text/notes/toolbar machinery is per-article. Per-segment selection is the safe, reachable solution.

## Testing

- `npm --prefix frontend test` — 256 tests green (17 new)
- `npx vite build` — compiles
- Browser smoke (against the built `dist/` with the real `besluit_zorgverzekering`, since the preview host has a TLS mismatch that blocks the sandbox browser):
  - Library: 354 entries → 62 article items, no leaf items; `Artikel 1` shows 41 segments merged; deep-link `2.4.1.a.1°` → `2.4`.
  - Editor: `Artikel 1` / `Artikel 2.4` each one tab; part selector with 41 resp. 15 segments; switching parts reuses the tab (no spam).

No custom CSS — everything via `ndd-*` components.

### DIFF
--- a/frontend/src/EditorApp.vue
+++ b/frontend/src/EditorApp.vue
@@ -13,6 +13,7 @@ import { useNotes, useResolvedDraftNotes } from './composables/useNotes.js';
 import { useDraftNotes } from './composables/useDraftNotes.js';
 import { useColorScheme } from './composables/useColorScheme.js';
 import { lastLibraryPath, sectionTarget } from './composables/useLastVisitedRoute.js';
+import { groupArticles, findGroupForArticleNumber } from './utils/articleGroups.js';
 import { SUPPORT_EMAIL } from './constants.js';
 import ArticleText from './components/ArticleText.vue';
 import AnnotatedText from './components/AnnotatedText.vue';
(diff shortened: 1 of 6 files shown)


tdjager:
## Browser smoke-test ✅

De preview-host (`editor.pr736.…`) heeft een TLS CN-mismatch die de sandbox-browser blokkeert, dus getest tegen de **gebouwde `dist/`** met de **echte** `besluit_zorgverzekering` (`2026-01-01`, 354 `articles[]`-entries) als gemockte corpus-API.

- **Kolom 2:** 354 entries → **62 artikel-items** (`Artikel 1`, `Artikel 2.1`, `Artikel 2.4`, …). Geen enkel blad-item meer (`1.e.1°`, `2.4.1.a.1°` weg).
- **Kolom 3:** klik op `Artikel 1` → URL `/…/1`, alle **41 segmenten** als één doorlopend artikel (definities `wet`, `verblijf`, … vloeien door).
- **Deep-link:** `/…/2.4.1.a.1°` normaliseert naar `/…/2.4`.
- **Machine/YAML-tab:** lege staat (deze wet heeft geen `machine_readable`) — geen crash. Per-segment provenance gedekt door `ArticleGroup`-unittests.

Unit + build: 251 tests groen (17 nieuw), `vite build` compileert.

tdjager:
Superseded by #743. We're fixing the over-splitting at the source — the harvester now defaults to **article-level granularity** (leden/onderdelen folded inline with their enumerators), so the UI can keep rendering `articles[]` 1:1 instead of aggregating segments and re-splitting. The split engine still supports finer granularity as an explicit choice; only the harvester default changed. Closing this in favour of #743.

PR #743, opened 2026-06-04
feat(harvester): article-level granularity by default

tdjager:
## What & why

The harvester split laws at every structural split point (`artikel`, `lid`, `li`), emitting one `articles[]` entry per lid/onderdeel/sub. `besluit_zorgverzekering` became **354 entries** (`1.a`, `1.e.1°`, `2.4.1.a.1°`, …), which exploded the editor's article list and dropped the onderdeel enumerators when the article was presented as a whole.

The deep split is **semantically inert**: the engine collects `machine_readable` across `articles[]` flatly, cross-law references target outputs/whole articles (zero sub-part refs in the corpus), and notes anchor by quoted text — none depend on the fine split. RFC-001 also states granularity is the author's choice.

So this makes **article-level the harvester default**: one component per `artikel`, with leden/onderdelen folded inline as blank-line-separated blocks that keep their enumerators (`1.`, `a.`, `1°`), matching how whole-article laws (e.g. `zorgverzekeringswet`) are already stored — so the existing markdown rendering shows the structure with no UI changes.

The leaf-splitting engine is **retained** (`SplitEngine::with_article_level`, off by default), so finer granularity stays available as an explicit choice; only the harvester's default is switched.

## Changes

- `splitting/engine.rs`: `with_article_level` + `extract_article_level_content` + `collect_article_blocks` (folds lid/onderdeel content with enumerator prefixes; marked lists → blocks, unmarked/dash/bullet lists stay inline; `redactie`/`tussenkop`/`meta-data` skipped). Article-level branch in `split()`. New unit tests.
- `harvester.rs`, `cvdr/parse.rs`: enable article-level at the harvest entry points; two CVDR tests updated to the article-level expectation.

## Verification

- Harvested `besluit_zorgverzekering` (BWBR0018492, 2026-01-01) end-to-end → **62 articles (was 354)**, schema-valid (v0.5.2). Article 1 is one entry with `a.`–`pp.` onderdelen and nested `1°.`–`4°.` inline, reference links and italic definition terms preserved.
- `cargo test` (harvester): 215 lib tests pass; `cargo fmt --check` + `clippy` clean.

## Follow-up (not in this PR)

- **Re-harvest the corpus** (`MinBZK/regelrecht-corpus`) so existing over-split laws become article-level. Audit first for any split law with hand-authored `machine_readable` on sub-segments (merge text + reattach the execution block); `besluit_zorgverzekering` has none.

## Supersedes

PR #736 (UI-side aggregation of split segments) — that approach is dropped in favour of fixing the granularity at the source; the UI keeps rendering `articles[]` 1:1.

### DIFF
--- a/packages/harvester/src/cvdr/parse.rs
+++ b/packages/harvester/src/cvdr/parse.rs
@@ -52,9 +52,11 @@ pub fn parse_cvdr_articles(xml: &str, base_url: &str) -> Result<ParsedCvdrConten
     let mut articles = Vec::new();
     let mut all_warnings: Vec<String> = Vec::new();
 
-    // Create split engine (reuses BWB article splitting logic)
+    // Create split engine (reuses BWB article splitting logic). Article-level
+    // granularity is the default: one entry per artikel with leden/onderdelen
+    // folded inline (enumerators kept).
     let hierarchy = create_dutch_law_hierarchy();
-    let engine = SplitEngine::new(hierarchy, LeafSplitStrategy);
+    let engine = SplitEngine::new(hierarchy, LeafSplitStrategy).with_article_level(true);
 
     // Find all artikel elements regardless of their position in the tree
     // CVDR structure varies: may be under body, hoofdstuk, paragraaf, afdeling, etc.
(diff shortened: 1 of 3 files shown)


tdjager:
Re: multi-`al` lids — by design, not a code change. A lid's continuation paragraphs carry no enumerator in the BWB source either (only the lid has a `lidnr`), so prefixing them would be incorrect; rendering them as unmarked paragraphs under the lid marker matches both wetten.overheid.nl and how whole-article laws are already stored in the corpus. Noted for the re-harvest audit as you suggested. No other findings — thanks for the review.

PR #746, opened 2026-06-04
feat(editor): search the full traject corpus, grouped by source

tdjager:
## What

The bibliotheek search box could only find the ~24 laws in a static favorites allowlist, because a traject loaded only those. Laws in a traject's **own private repo** (e.g. `MinBZK/regelrecht-corpus-BES`) or anywhere in the central corpus were unfindable. This makes the search reach **every** law across all of a traject's sources, grouped by source.

## How it works (final design)

The naive version (eagerly fetching every law's body) was too heavy — the central corpus is the full BWB corpus (thousands of laws), so per-file Contents API calls timed out and tripped GitHub's rate limit. Instead:

1. **Lazy index** — opening a traject enumerates every law via the **Trees API (1 call per source)**, building a metadata-only index (`law_id` from the directory name, source, path) with **no bodies fetched**. A law's YAML is fetched lazily on first open (`law_yaml` -> backend `read_file`, 1 call). Cold index build is about 2.6 s once per pod; warm reads are instant.
2. **Server-side search** — `/corpus/laws?q=<term>` filters the full index on the backend (matching `law_id`/name, underscores treated as spaces), orders by source priority (private repo first) then `law_id`, capped to the limit. The popover queries this (debounced, stale-response-guarded) instead of filtering a client-loaded page — so search reaches the whole corpus, not just the first 1000 rows.
3. **Grouped UI** — results render under one header per source, the traject's own repo (priority 0) first, then central. The wetten.overheid.nl (BWB) fallback fires only on a genuine corpus miss.

## Verified on the preview (pr746)

| Query | Result |
|---|---|
| `zorgverzekering` | BES group first (Besluit Zorgverzekering Bes, ...) then Centrale corpus — grouped, private on top |
| `bes` | BES (23) first, then Centrale (37) |
| `kieswet` | found (was invisible past the old client page cap) |
| `wegenverkeerswet` | found in Centrale — no false external fallback |
| `qzxqzx-not-a-law` | 0 internal -> wetten.overheid.nl fallback |

Index/search timing: ~2.6 s cold index build per pod, ~28 ms per query after.

## Notes

- UI uses only MinBZK design-system `nldd-*` components (`nldd-title`/`nldd-list`/`nldd-spacer`/`nldd-inline-dialog`). **No custom CSS.**
- Search list shows the humanized `law_id` until a law is opened (real names live in the body); fine for search-by-name.
- Three pre-existing editor-api integration test files are stale against the current trajects API (unrelated to this PR, not compiled by CI); only their reference to the removed `AppState.favorites` field was cleaned up.

## Tests
- Rust: build + clippy clean; corpus/editor unit tests green (incl. a new metadata-entry test).
- Frontend: vite build + `SearchPopover.test.js` (server-side query, grouping order, debounce).

### DIFF
--- a/frontend/src/EditorApp.vue
+++ b/frontend/src/EditorApp.vue
@@ -1882,7 +1882,6 @@ async function handleActionSave() {
   <TrajectDocuments />
   <SearchPopover
     ref="searchPopoverRef"
-    :laws="corpusLaws"
     @select-law="onSearchSelectLaw"
     @harvest-available="onSearchHarvestAvailable"
   />
--- a/frontend/src/LibraryApp.vue
+++ b/frontend/src/LibraryApp.vue
@@ -144,17 +144,19 @@ const detailView = computed({
 const activeAction = ref(null);
 
 // Curated sidebar sections (in render order). Each entry is
-// `{ key, title, laws }`; a `null` title renders a headerless list (the
-// no-favorites fallback). Empty sections are never pushed, so the template
+// `{ key, title, laws }`. Empty sections are never pushed, so the template
 // can iterate without per-section emptiness checks.
 //
 //   - "Bewerkt in dit traject" comes first: it's the small, high-signal,
-//     context-specific set, so it sits above the larger favorites/full list.
+//     context-specific set, so it sits above favorites.
 //     Only present when a traject is active and the diff is non-empty.
-//   - "Favorieten": the user's personal favorites. When the user has none we
-//     fall back to the full corpus (today's behavior) under no heading,
-//     rather than showing an empty panel — full browse otherwise lives in
-//     the search popover.
+//   - "Favorieten": the user's personal favorites.
+//
+// There is deliberately NO full-corpus fallback: the central corpus is the
+// full BWB corpus (thousands of laws), so dumping it into the sidebar isn't
+// useful and is exactly the "huge pile" we don't want loaded here. When
+// nothing is curated yet, the template shows a search CTA instead — full
+// browse lives in the search popover.
 const sidebarSections = computed(() => {
   const list = laws.value;
   const sections = [];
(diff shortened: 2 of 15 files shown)


tdjager:
Thanks for the review — addressed in 08895b9:

- 🟠 **Search clear / stale results**: the clear branch now bumps `searchSeq`, so an in-flight corpus fetch is discarded on resolve instead of repopulating the cleared box or firing a spurious harvest search. Added a regression test.
- 🟡 **`law_yaml` re-fetch**: the lazily-fetched body is now cached in the overlay, so re-reads of an unloaded law (read-only views, reloads, a second tab) don't each spend another Contents API call.
- 🟡 **`is_loaded()` empty-content sentinel**: left as-is. The overlay cache above short-circuits before the `is_loaded` check, so even a genuinely-empty body is a one-shot fetch — the repeated-fetch symptom is gone without refactoring `yaml_content` to `Option<String>` (which touches every read site for an unlikely case).

tdjager:
Round 2 addressed in 24cc022:

- 🟠 **Backend error → external fallback**: a non-OK corpus response now sets a `searchFailed` flag and shows an error in the popover instead of collapsing to `[]` and cascading to the harvest search. The external fallback only fires on a genuine empty corpus result. Added a regression test.
- 🟠 **Writable-own "escalated" comment**: corrected to match behaviour. I kept the **degrade** (log at error, traject still opens with those laws missing from search until rebuild) rather than returning `Err` — failing the whole traject on a transient index-enumeration hiccup is worse than a temporarily-incomplete search. The genuine fail-closed path is the writable-own *backend* init above, which already returns `Err`. Comment now says so.
- 🟡 **`ids`/`q` precedence**: documented on `PaginationParams.ids` that `ids` takes precedence and `q` is ignored when both are sent.

tdjager:
Round 3 addressed in 57b3c7b:

- 🟠 **`law_yaml` masking backend errors as 404**: `law_yaml` now returns `Result<Option<String>>`. A lazy-fetch read error (GitHub throttle/429, token expiry, network) is propagated and surfaced as a logged **502 "kon wet niet laden"** via the new `read_law_yaml` helper, instead of the `.ok().flatten()` that collapsed it to `None`/404. A genuine miss (not indexed, missing file, backend-skipped source) stays `Ok(None)` → 404.
- 🟡 **`is_loaded()` sentinel**: added a doc note that empty-`yaml_content` is a safe "unloaded" sentinel only because corpus YAML is never legitimately empty (a genuinely-empty file is re-fetched once, then the overlay short-circuits).

PR #762, opened 2026-06-05
fix(editor): stop scenario dependency-scan request storm + un-gate result buttons

tdjager:
## Problem

The editor's scenario panel hung on **"Afhankelijkheden laden"** and the **Resultaat / Graaf** buttons stayed greyed out (not clickable). Reproduced on a traject pointing at a private corpus repo (zorgtoeslagwet, art. 2):

Opening the panel for **one** law fired **~1016 requests** and never settled:
- **258 unique laws** fetched — the traject federates the whole central corpus, and the client-side `discoverImplementors` fetched+parsed *every* law YAML over HTTP to find implementing regulations.
- **~4× per law** — the `ScenarioBuilder` deps watch re-fired on load (`lawYaml`/`ready`/`formState` settle on separate ticks), restarting the whole scan before it could finish, so `depsReady` never became true.
- The backend (GitHub-backed, rate-limited) returned some **502s** under the load.
- With `depsReady` never true, scenarios never auto-executed → no cached result → the **Resultaat/Graaf buttons were `:disabled`**, which also contradicts `onShowDetails`'s own "don't be a dead gate" design.

## Fix

- **Server-side implementors endpoint (B).** New `GET /api/corpus/laws/{id}/implementors` (+ traject variant) scans the in-memory corpus for laws whose `machine_readable.implements` references `{id}`. Resolves each candidate body through the read scope so a traject's lazily-indexed federated laws are fetched + cached (not read as empty snapshot bodies). The editor now makes **one** request instead of fetching every law.
- **Debounced deps cascade (C).** `ScenarioBuilder`'s dependency watch coalesces the load-time burst into a single run, so the scan isn't started-and-abandoned 4×.
- **Un-gated buttons (A).** Removed `:disabled` on Resultaat/Graaf — `onShowDetails` already opens the sheet and lets it handle loading/empty/reload, so the user can always open the trace/graph.

## Verification

- Backend unit tests for `collect_law_implements` (parsing + dedup + malformed-tolerance).
- Live on a local stack: `GET /api/corpus/laws/zorgtoeslagwet/implementors` → `["regeling_standaardpremie"]`; `kieswet/implementors` → `[]`.
- Frontend: new `useDependencies.test.js` asserts a single implementors request (no `?limit=1000` corpus scan) and that implementor + source.regulation deps load; full suite green (266).
- End-to-end on the PR preview against the reproduction traject is the final check (see Test plan).

## Test plan

- [ ] On the preview, open a traject (custom repo) → zorgtoeslagwet art. 2 → Scenario's: panel loads quickly (no perpetual "Afhankelijkheden laden"), and **Resultaat/Graaf are clickable**.
- [ ] Network shows a single `/implementors` call, not a per-law storm.

## Notes / follow-ups

- A backend integration test for the traject lazy-load path (metadata-only entries) isn't included — it needs a constructed traject scope; covered for now by the unit test + live preview verification.
- The global `implementors` endpoint reads bodies via the scope too; global non-loaded GitHub laws (favorites-only init) could still be missed, same root cause, lower impact (local corpus is body-loaded). Follow-up: a build-time `implements` index would make this O(1) and fully eager.

### DIFF
--- a/frontend/src/components/ScenarioBuilder.vue
+++ b/frontend/src/components/ScenarioBuilder.vue
@@ -181,39 +181,57 @@ watch(() => props.lawYaml, (val, prev) => {
 
 onBeforeUnmount(() => clearTimeout(lawYamlDebounce));
 
+// Run the dependency cascade for the current law + scenarios. Reads the
+// latest prop/state values at call time (not captured watch args) so a
+// debounced run always uses the freshest inputs.
+async function runDependencyLoad() {
+  const lawYaml = debouncedLawYaml.value;
+  if (!lawYaml || !props.ready || !props.engine) return;
+  const version = ++watchVersion;
+  depsReady.value = false;
+
+  await loadAllDependencies(lawYaml, props.engine, fetchLawYaml, props.trajectRef);
+  if (version !== watchVersion) return;
+
+  // Also load dependencies from scenario background + per-scenario steps
+  if (formState.value) {
+    const allDeps = new Set();
+    for (const dep of formState.value.background?.dependencies || []) allDeps.add(dep);
+    for (const sc of formState.value.scenarios || []) {
+      for (const dep of sc.setup?.dependencies || []) allDeps.add(dep);
+    }
+    for (const depId of allDeps) {
+      try {
+        if (!props.engine.hasLaw(depId)) {
+          const yaml = await fetchLawYaml(depId);
+          props.engine.loadLaw(yaml);
+        }
+      } catch (e) {
+        console.warn(`Failed to load scenario dependency '${depId}':`, e);
+      }
+    }
+  }
+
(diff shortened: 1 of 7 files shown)


tdjager:
Thanks — both points addressed:

1. **Sequential `scope.law_yaml` in `implementors_in_scope`**: acknowledged and intentional for this PR. The scan is off the critical path, overlay-cached after first run, and guarded once per (traject, law). The proper O(1) fix — a build-time `implements` reverse-index so no body fetch is needed — is tracked as a follow-up in the PR description.
2. **`loadImplementors` has no AbortController**: added an inline comment documenting the deliberate fire-and-forget and why it's safe (Vue ignores post-unmount ref writes, the WASM engine is shared, the guard resets on error).

PR #781, opened 2026-06-10
feat(frontend): persistent AppShell so editor↔library keeps the chrome

tdjager:
## Why

The editor and library are already a single Vue SPA, but `App.vue` was a bare `<router-view>` mapping `/editor` and `/library` to two independent top-level components. Each Bibliotheek↔Editor tab switch unmounted one whole component tree and mounted the other, **rebuilding the shared chrome** (the three responsive toolbars, tab-bar, settings menu, traject menu). The document never actually reloaded, but the visible chrome rebuild reads as a full refresh.

## What

- **`AppShell.vue`** (new) — a persistent parent route that owns the shared chrome (`nldd-app-view` → `nldd-bar-split-view` + the md/lg/sm toolbars: tab-bar, search trigger, `TrajectMenu`, settings menu) and a nested `<router-view>`. It also mounts `TrajectDocuments` once.
- **`EditorView.vue` / `LibraryView.vue`** (renamed from `*App.vue`) — now children of `AppShell`. They lose their chrome and keep only their body, rendered through the shell's nested router-view. A switch swaps only that nested view, so the shell instance is reused — **the chrome never rebuilds**.
- **`useAppChrome.js`** (new) — a small module-level chrome store. The active view publishes its per-view toolbar extras: each view registers its `SearchPopover`; the editor additionally publishes the federated "PR #N" indicator and the open document tabs (cleared on unmount). The shell's `document-tabs` pane renders only while there are tabs, so the library never shows an empty tab bar.
- **Routing** — editor/library are nested under one `AppShell` record. Route names, paths and redirects are unchanged, so deep links and the traject-aware `sectionTarget` logic keep working.

## Design note — chrome store instead of `<Teleport>`

The [design doc](docs/superpowers/specs/2026-06-09-shared-editor-library-shell-design.md) originally chose `<Teleport>`. During implementation that proved unworkable: `nldd-toolbar` only projects slotted items that are its *direct* children, so a teleport target inside the toolbar breaks projection, and teleporting into the toolbar host appends items *after* the shell's own `slot="end"` items — reordering the bar (search/PR would land after the settings menu). The chrome store lets the shell render every item in the original order, guaranteeing visual parity. The doc is updated to reflect this.

Two parity details surfaced while implementing (both noted in the doc):
- Since PR #746 the editor also has a search field, so search is shared chrome (not library-only as the doc's snapshot said).
- The settings menu now lists the full **7-flag** editor set; the library previously listed only 4. The extra flags only affect editor panes, so exposing them from the library is harmless.
- The shared tab-bar standardises on the library's `navigation` attribute, so the editor's section tab-bar (previously a plain tablist) is now also a `nav` landmark — an intentional unification since both tabs perform cross-section navigation. Verified no visual change in Playwright.

## Verification

- `npm run test` — **268 unit tests pass** (incl. `router.test.js`).
- `npm run build` — succeeds; `EditorView` stays lazily code-split.
- Playwright round-trip on a running dev server:
  - Marked a shell DOM node, switched Bibliotheek→Editor→Bibliotheek — the **same node survived both switches** (no remount).
  - `document-tabs` pane appears on the editor (store-driven) and clears on return to the library.
  - Settings menu shows all 7 flags; typing in the lg search field opens the popover (store wiring intact).

Editor view after the switch — the toolbar/chrome is the same persistent shell node, with the editor's document-tabs published into it:

editor with persistent chrome and document tabs

## Out of scope

`<keep-alive>` / body-state preservation across switches (view state is URL-driven and reloads from the URL on remount), toolbar redesign.


tdjager:
Re: the `navigation` attribute on the editor tab-bar — this is an intentional, conscious unification, not an accidental side effect. Both tabs in the shared shell perform cross-section navigation (Bibliotheek↔Editor), so the section tab-bar is correctly a `nav` landmark; the old editor's plain tablist (without `navigation`) was the inconsistency, and the shell standardises on the library's `navigation` semantics. Verified in Playwright that the editor toolbar renders identically (the tab-bar exposes as `navigation "Tabs"` with no visual difference). Keeping it as-is.

tdjager:
Re: review finding **"race on editor→editor-traject navigation"** — disputed after empirical verification on the pr781 preview.

The premise ("Vue Router creates a new EditorView instance") doesn't hold: both `editor` and `editor-traject` records lazy-resolve to the *same* component object, and the nested `<router-view>` has no `:key`, so Vue reuses the existing instance on that navigation. No unmount happens, so `onBeforeUnmount(clearEditorChrome)` never fires and the described register/clear race can't occur. (Same topology as on main, where the root was a bare `<router-view/>`.)

Verified in the preview deployment: opened `/editor/zorgtoeslagwet/2` (document tab visible) → picked a traject via TrajectMenu → landed on `/editor/zorgtoeslag-private-repo-…/zorgtoeslagwet/2` with the document-tab-bar intact. Also checked the editor↔library roundtrip: tab bar hidden in the library, restored on return.

The 🟡 minor (composable return exposing `openSearch`/`onBarSearchKeydown` that no consumer uses) was valid — fixed in 26c18b8.

PR #791, opened 2026-06-11
fix(engine): round eurocent outputs only at the outermost result boundary

tdjager:
## Problem

TypeSpec `eurocent` → integer rounding lived inside `evaluate_article_with_service`, which executes at **every** article boundary — same-law internal references, cross-law resolutions, open-term resolutions, hooks and overrides — not only at the top-level API output as its comment claimed.

That means an intermediate eurocent value produced by a sub-law was rounded at the sub-law boundary, then combined further by the calling law and **rounded again** at the outer boundary. Double rounding can shift a legally binding amount by a cent.

### Concrete example (the regression test)

- Sub-law: `deelbedrag = 199 × 0.5 = 99.5` cent
- Parent law: `totaal = 2 × deelbedrag` → mathematically exactly **199** cent

Old behavior: 99.5 was rounded to 100 at the sub-law boundary, so the parent returned **200** cent — one cent too much. Same failure existed for same-law internal references (`source.output`).

## Fix

- Removed the rounding from `evaluate_article_with_service` (and replaced the inaccurate comment with one explaining why rounding must NOT happen there).
- Added `round_eurocent_outputs`, applied exactly once at the public API boundaries:
  - `evaluate_law_multi_internal` (covers `evaluate_law`, `evaluate_law_output`, the `with_trace` variants and `evaluate_uri`)
  - `evaluate_law_output_with_shared_trace` (stage fall-through for non-procedure laws)
  - `execute_stage_internal` on `ExecutionOutcome::Complete` (final accumulated stage result)
- The producing article per output is looked up via `output_provenance` (Direct/Reactive/Override), so hook and override outputs from *other* laws keep their eurocent rounding; outputs without provenance fall back to the requested law's output index.
- Intermediate values now flow unrounded as `Float`; arithmetic ops already promote to `Float` when any operand is a float, so precision is preserved up to the boundary. TypeSpec metadata handling is unchanged.

Notes on intentional behavior details:
- `ExecutionOutcome::Yielded` outputs are intermediate stage state (they feed the resumed computation), so they are deliberately **not** rounded; rounding happens once on `Complete`.
- The per-resolution cache stores unrounded values — correct, since cached entries are consumed by nested resolution.

## Tests

- `test_eurocent_rounds_once_at_outer_boundary_cross_law` — fails on the old code (returns 200, expects 199)
- `test_eurocent_rounds_once_at_outer_boundary_internal_reference` — fails on the old code (200 vs 199)
- `test_eurocent_rounds_at_top_level_boundary` — verifies the boundary rounding itself still works (99.5 → 100)

Both regression tests were verified to fail against the old code before applying the fix.

## Verification

- `cargo test -p regelrecht-engine`: all green (332 lib tests + integration suites, incl. golden tests)
- `just bdd`: **9 features, 66/66 scenarios, 390/390 steps passed** — no BDD expectation needed changing; all legislature-intended example outcomes (e.g. zorgtoeslag €2107.26) are unchanged
- `just format` / `just lint`: clean

## Follow-up

This PR removes the only intermediate-rounding mechanism the engine had. Laws whose **text** mandates rounding at an intermediate step (e.g. "rekenkundig afgerond op hele eurocenten" on a sub-amount) can no longer express that: they would need an explicit `ROUND` operation in the engine's op set so the law itself states where rounding legally happens. Decision recorded here as the follow-up; the current corpus is unaffected (BDD 66/66 unchanged).

## Note for frontend

Trace nodes and `resolved_inputs` now show **unrounded intermediate floats** (e.g. `99.5` cent) where they previously showed rounded integers. This is truthful — it is the actual value the calculation continues with — but visibly different in trace/debug views. Only the final boundary outputs are rounded integers.

## Update: generalized to `type_spec.precision`
The eurocent special-case is gone: boundary rounding is now driven by the schema's existing `type_spec.precision` field (which the engine previously parsed-and-dropped — only `unit` was deserialized). Any output with `precision: N` rounds half-away-from-zero to N decimals at the public boundary; `precision: 0` converts to an integer. `unit: eurocent` defaults to precision 0, so all existing behavior (incl. BDD 66/66) is byte-identical. Rounding that a law's *text* prescribes at an intermediate step is deliberately not config — that will be an explicit `ROUND` operation (see Follow-up above).

### DIFF
--- a/packages/engine/src/article.rs
+++ b/packages/engine/src/article.rs
@@ -39,14 +39,16 @@ pub struct LegalBasis {
 }
 
 /// Type specification for input/output fields.
-///
-/// Currently only contains unit specification, but may be extended
-/// with additional type metadata (precision, range, format) as the schema evolves.
 #[derive(Debug, Clone, PartialEq, Default, Serialize, Deserialize)]
 pub struct TypeSpec {
     /// Unit of measurement (e.g., "eurocent", "days", "percentage")
     #[serde(default)]
     pub unit: Option<String>,
+    /// Number of decimal places for numeric values (schema:
+    /// `type_spec.precision`). Drives boundary rounding of outputs;
+    /// `unit: eurocent` defaults to a precision of 0 when absent.
+    #[serde(default)]
+    pub precision: Option<u32>,
 }
 
 /// Source specification for input fields
--- a/packages/engine/src/engine.rs
+++ b/packages/engine/src/engine.rs
@@ -31,7 +31,11 @@ use std::collections::{BTreeMap, HashSet};
 use std::rc::Rc;
 
 /// Provenance of an output value: how it was produced during execution.
-#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize)]
+///
+/// `Deserialize` is needed because provenance is carried inside the
+/// serializable [`StageState`](crate::service::StageState) across
+/// yield/resume cycles.
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 #[serde(tag = "type")]
 pub enum OutputProvenance {
     /// Produced by the article's own actions.
--- a/packages/engine/src/service.rs
+++ b/packages/engine/src/service.rs
@@ -333,6 +333,16 @@ pub struct StageState {
     pub current_stage: String,
     /// Outputs accumulated from all completed stages
     pub accumulated_outputs: BTreeMap<String, Value>,
+    /// Per-output provenance accumulated from all completed stages.
+    ///
+    /// Needed so an output produced in an *earlier* stage (e.g. a Reactive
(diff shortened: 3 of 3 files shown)


tdjager:
Closing as **superseded by RFC-024** (exact-decimal arithmetic + rounding as an explicit ROUND/CEIL/FLOOR/TRUNCATE operation, merged in #848).

This PR took the opposite approach — *automatic* eurocent rounding at the outermost result boundary. RFC-024 establishes that rounding must be modeled explicitly in the law (never inferred from a unit), and the engine now uses `Value::Decimal` for exact arithmetic, so boundary auto-rounding is no longer the right mechanism. The branch is also conflicting with main. The underlying audit concern (advice 6: no double-rounding across nested article boundaries) is resolved by RFC-024's model rather than by this change.

PR #796, opened 2026-06-11
refactor: extract the law document model into a canonical leaf crate

tdjager:
## Summary

The law-YAML document model — the shape of the files under `corpus/regulation/` —
was implemented in several places because no crate could own it: the engine held
the full serde model, while `corpus`, `tui` and `pipeline` each re-derived the
parts they needed (the engine can't be a shared dependency for all of them). This
PR gives the model a single home.

## What changed

1. **New leaf crate `regelrecht-law-model`** — the one Rust representation of the
   law document model (the structs/enums from engine's `article.rs` plus
   `Value`/`Operation`/`ParameterType` from `types.rs`). Dependency-light
   (serde + serde_json, no engine/runtime deps). The engine re-exports everything
   at the historical `engine::article` / `engine::types` paths, so no downstream
   code changes. YAML loading with the security limits stays in the engine as a
   `LawLoad` extension trait (it depends on the engine's config + error type).
2. **One shared tolerant header parser** (`parse_law_header` / `LawHeader`) in the
   new crate, now used by both the **tui** (`corpus_scanner`) and **corpus**
   (`extract_law_id` / `extract_law_name`), which each carried their own line
   scanner. It matches header keys at column 0 only and tolerates a malformed
   body (editor drafts, mid-enrichment files).

The JSON schema under `schema/` stays the source of truth — no code generation is
introduced.

## Deliberately NOT collapsed

Several of the remaining "copies" are not redundant model definitions but
intentionally different representations; forcing them onto the strict model would
change behavior, so they are left as-is (and documented):

- **corpus body-view** (`collect_law_outputs` / `collect_law_implements` /
  `resolve_display_name`): a deliberately permissive view (every field optional)
  so it keeps working on incomplete editor drafts.
- **pipeline `enrich.rs`**: counts articles via a structured `serde_yaml_ng::Value`
  walk for accuracy — a different algorithm, not a model copy.
- **harvester `writer.rs`**: the harvest output format (raw articles with BWB
  `references`, pre-enrichment, no `machine_readable`) — a different shape.
- **frontend JS**: would require code generation; out of scope.

## No behavior / format changes

The law YAML on disk is unchanged; no schema version is introduced. engine (324)
+ tui + corpus + law-model unit tests pass; BDD 71 scenarios / 408 steps pass;
WASM build succeeds; `just format` and `just lint` (`-Dwarnings`) clean.


tdjager:
Re: `serde_json` is a hard dep in a crate described as "serde only" — corrected the crate doc wording to "serde + serde_json". Keeping `serde_json` as a plain (non-feature) dependency on purpose: the `Value` ↔ `serde_json::Value` conversions are part of the value type the engine relies on, so every real consumer of the model needs them — gating them behind a feature would just force all consumers to enable it for no benefit. The two minor nits (broken intra-doc link, missing blank line) are fixed.

tdjager:
Round 2 review responses:
- **article_count over-count**: fixed — the `LawHeader::article_count` doc now states it's a line-based `- number:` heuristic, not a structural count.
- **`publish = false`**: disputing — there is no such workspace convention. None of the existing crates (engine, shared, corpus, harvester, pipeline, …) set `publish`; they all carry the same crates.io metadata block. `law-model` matches them exactly, so adding `publish = false` would make it the odd one out rather than match convention.
- **TUI column-0 matching** and **`$schema` quote stripping**: both intentional (the column-0 fix is covered by the `ignores_indented_id` test; quote-stripping makes `$schema` consistent with every other field). No change needed.

PR #832, opened 2026-06-18
feat(editor): opt-in local test corpus trajecten

tdjager:
## Summary

The **local test corpus** currently shows up in law search everywhere — in global browse and in *every* traject — because traject creation seeds all registry sources, including `local`. This makes it opt-in instead.

- **Drop `local` from the shared `corpus-registry.yaml`.** With it gone from the registry it no longer appears in global browse or in newly created normal trajecten — no per-source filtering code needed. (The dead duplicate `frontend/public/corpus-registry.yaml` is removed; it is regenerated at `prebuild` from the root registry.)
- **Opt-in via a reserved repo name.** Creating a traject with the existing "custom repo" fields and owner `local` / repo `test-corpus` (case-insensitive) produces a **read-only local-test-corpus traject**: the writable-own source is a synthesized `local` source (`corpus/regulation/nl`), and it reads **local + minbzk-central** (so local laws that reference central laws still resolve). This is a **developer feature** — deliberately *not* surfaced in the UI (no create-form hint, no read-only banner) to avoid clutter.
- **Read-only enforcement.** A new `trajects.read_only` column is threaded into `TrajectCorpus`; the six traject write/delete handlers (`save_scenario`, `save_annotations`, `save_law`, `save_traject_document`, `delete_scenario`, `delete_traject_document`) reject writes with a 403. This prevents an edit from silently mutating the baked-in container copy of the corpus.

Applies to everyone (committed config/code, no per-machine setup). Existing trajecten are unchanged (no retroactive migration of their sources).

## Verified on the `pr832` preview

- Global browse (no traject): only `minbzk-central`, no local corpus.
- Reserved-name traject: `read_only: true`, sources = `Local Test Corpus` (local) + `minbzk-central`, local laws searchable inside it.
- Save against it → `403` (read-only).

## Reviewer notes

- **editor-api integration tests are not wired into CI** (pre-existing gap): the new `TestDb` tests in `trajects_test.rs` were verified locally against a real Postgres container, but `just test-all` / CI does not run the `editor-api-integration-test` recipe. The in-crate unit tests for the read-only guard *do* run in CI's `Test` job.
- The `trajects_test.rs` `set_active`/`leave` edits **repair pre-existing breakage on `main`** (the file referenced a removed `set_active` handler and an old `leave` signature, so it didn't compile — editor-api tests aren't compiled in CI, which is why it went unnoticed). No live coverage was removed.
- The read-only traject relies on the editor image baking `corpus/regulation/` at `/app` (confirmed in `frontend/Dockerfile`).

## Test plan

- [x] Global search (no traject) shows no local corpus results.
- [x] A normal traject's search shows no local laws.
- [x] A traject created with owner `local` / repo `test-corpus` shows local + central laws, is read-only, and saving returns 403.


tdjager:
Re: `build_traject_corpus` fetching `read_only` in a separate query — keeping it as-is, intentionally. The existing query returns *multiple* `traject_corpus_sources` rows (one per source, `ORDER BY priority`); folding in the single `trajects.read_only` scalar means a JOIN plus carrying a duplicated bool on every `TrajectSourceRow`, which reads worse than one extra PK lookup. `build_traject_corpus` runs on corpus init/refresh (cached), not per-request, so the round-trip cost is negligible, and the `fetch_optional(...).unwrap_or(false)` is deliberately defensive against a not-yet-migrated DB. Noted as a possible follow-up but not changing it here.

PR #848, opened 2026-06-19
feat(engine): rounding operations + exact decimal arithmetic (RFC-024)

tdjager:
## Summary

Implements the **engine track for [RFC-024](docs/src/content/rfcs/rfc-024.md) (Precision and Rounding)** plus the exact-decimal foundation it depends on. This is PR1 of two; PR2 will implement RFC-023 (Units).

Both RFCs (merged in #823) are deliberately *silent on the engine* and decide only the YAML/modeling vocabulary — this PR builds the engine behaviour.

## What's in it

**Exact decimal arithmetic**
- `Value::Float(f64)` → `Value::Decimal(rust_decimal::Decimal)` across the engine, so intermediate calculations carry full precision and rounding only ever happens where a law asks for it.
- Non-finite `f64` at the boundary maps to `Null`; `DIVIDE`-by-zero still errors. Behaviour-preserving — the only output change is that f64 representation noise (`1508.2111200000002`) no longer appears (`1508.21112`).

**Four rounding operations (RFC-024)**
- `ROUND` / `CEIL` / `FLOOR` / `TRUNCATE`, each a unary `value:` + a `precision:` (decimal places, in the value's own unit). `ROUND` is rekenkundig/half-up; `precision: 0` = whole units, negative precision rounds to tens/hundreds (e.g. `-2` = whole euros on a eurocent value).
- Available both nested (`value:`) and inline (action-level `operation:`). Negative-precision scaling uses checked arithmetic so extreme magnitudes error rather than panic.

**No more implicit money rounding**
- Removed the engine's implicit eurocent boundary rounding. Money is no longer rounded behind the law's back — rounding is a law-modeled instruction.
- Migrated `wet_op_de_zorgtoeslag` (reference law) to an explicit `ROUND`, preserving the legislature-intended outcome (€2096.92).

**Schema v0.5.5**
- `roundingOperation` definition, `precision` on actions, the four ops in `operationType` + the `operation` oneOf; repointed `schema/latest`; updated `validate.rs`, `config::SUPPORTED_SCHEMAS`, Cargo metadata and the conformance manifest.
- Documented rounding/precision in the law-format guide.

## Verification

`just format` · `just lint` · `just build-check` (all packages) · `just validate` (whole corpus) all green. Engine test suite: **370 unit + 14 law-model + 12 golden + 4 trace + 3 conformance + 76 BDD scenarios** green.

## Out of scope (deliberate)

- **Corpus rounding audit** — AWIR/zvw "hele euro's" → explicit `ROUND` and the `test_untranslatables` "naar boven afgerond" → `CEIL` flip are legally sensitive per-law calls; the corpus does not depend on this change, so they're left for a separate pass.
- **RFC-023 (Units)** — schema `euro`/`ratio`/`percentage` + unit algebra/mismatch engine, as PR2.

## Notes for reviewers

- `Value::Decimal` currently serializes to JSON via `f64` at the API edge (unchanged from before); preserving exactness end-to-end to the frontend would be a separate, coordinated decision.


tdjager:
Re: `rust_decimal` default features pulling in `rkyv`/`bitvec`/`borsh` etc. — disputing this one: those crates are **not in the compiled dependency tree**. `cargo tree -e normal -i rkyv` reports "nothing to print" and `cargo tree -e normal | grep -c rkyv` is `0`; they only appear as unused optional entries in `Cargo.lock`, which are never built and don't affect compile time. `rust_decimal`'s default features here resolve to just `serde` + `std` (both required — `Value::Decimal` serializes to JSON and we use std), so `default-features = false` + re-adding them would compile the same set with no benefit and added risk. Leaving as-is. (Fixed the other three findings.)

packages/Cargo.toml
```diff
@@ -32,6 +32,8 @@ regex = "1.11"
 # while corpus/harvester/pipeline use 0.13. Aligning them would require
 # upgrading openidconnect, which is out of scope for this cleanup.
 roxmltree = "0.21"
```
tdjager, line 34: I made the features explicit (`default-features = false, features = ["maths", "serde", "std"]`) — good practice regardless. But for the record, `rkyv`/`borsh`/`bitvec` are **not compiled into the WASM binary** (or any build): `cargo tree -i rkyv --target all -e all` reports "nothing to print" and `cargo tree -e normal --target all | grep -c rkyv` is `0`, both before and after this change. They remain as resolved-but-unused entries in Cargo.lock (rust_decimal lists them as *optional* deps, so cargo records them in the resolution graph), but no feature enables them, so they never reach the cdylib. So there was no WASM bloat — the change just makes our feature set explicit.

PR #849, opened 2026-06-19
fix(engine): load all law versions for date-aware cross-law resolution

tdjager:
## Problem

Running a `wet_op_de_zorgtoeslag` scenario as of `2025-01-01` in a federated
(local + minbzk-central) test-corpus traject fails:

```
Execution failed: No version of law 'zorgverzekeringswet' in force on 2025-01-01 (not yet in force)
```

## Root cause

Scenario execution runs in-browser via WASM. The dependency walker fetched
**one** YAML per law id via `/api/.../corpus/laws/{id}`, and the backend
`SourceMap` collapses each `$id` to a single **"best for today"** version
(`pick_best_version` uses *today's* date, not the scenario's `calculation_date`).
So when a referenced law (`zorgverzekeringswet`) has a future-dated version
(central harvested a `2026-01-01`), the engine received only that future version
and a past-dated scenario failed `NotYetInForce`.

The engine's `select_in` is correct *given all versions* (a `valid_from`-less or
in-force version is matched); it simply never saw more than one version.

## Fix

Feed **every** version of each referenced law to the engine and let its
date-aware `select_in` pick the one in force on the scenario's calculation date.

- **corpus** (`source_map.rs`): `SourceMap` now also retains all versions per
  `$id` (`versions` map, populated in `insert`), unioned across sources with
  per-date priority resolution; new `get_law_versions`. The single-best `laws`
  map / `get_law` are unchanged, so existing single-version callers are
  untouched.
- **editor-api** (`traject_corpus.rs`, `corpus_handlers.rs`, `main.rs`): new
  `GET /api/corpus/laws/{id}/versions` and the traject-scoped variant returning
  a JSON array of every version body (lazy fetch + read-your-writes overlay for
  the saved version).
- **frontend** (`corpusUrls.js`, `useDependencies.js`, `useEngine.js`,
  `ScenarioBuilder.vue`): the dependency walker, scenario-step loader,
  implementors loader and `loadDependency` now load every version into the WASM
  engine (it keys versions by `($id, valid_from)`, so they coexist).

## Tests (target the execution version-loading path, not `select_in` alone)

- **engine** (`service.rs`): execute a law referencing another that has both an
  in-force and a future-dated version, as of a date before the future one →
  resolves the in-force version; plus a **control** asserting that with *only*
  the future version loaded the same execution fails `NotYetInForce` (proving
  the fix is the version *set* reaching the engine).
- **corpus** (`source_map.rs`): `get_law_versions` returns the cross-source
  union (in-force + future), order-independent.
- **frontend** (`useDependencies.test.js`): loads *every* version of a
  dependency; empty version list is routed to harvest (not silently dropped).

All green: `just check` (format + lint + build-check + validate + test) and the
full frontend vitest suite (356 tests).

## Corpus-data note (investigated, intentionally not changed)

The data oddities are legitimate, not bugs: local `zorgverzekeringswet/2025-01-01.yaml`
has **no** `valid_from` (always in force — one of a few such local laws), and
central legitimately harvested a future `2026-01-01` version. Editing one law's
YAML would only mask the general defect; the fix makes execution date-aware
regardless of which source/version wins.

### DIFF
--- a/frontend/src/components/ScenarioBuilder.vue
+++ b/frontend/src/components/ScenarioBuilder.vue
@@ -169,7 +169,11 @@ async function fetchLawVersions(lawId) {
     errorMessage: (status) => `Failed to fetch versions of law '${lawId}': ${status}`,
   });
   const list = Array.isArray(yamls) ? yamls : [];
-  versionsCache[lawId] = list;
+  // Only cache a non-empty result. An empty array (unknown/not-yet-harvested
+  // law) must stay uncached so a retry after harvest re-fetches rather than
+  // returning the stale `[]` — `[]` is truthy, so caching it would short-
+  // circuit the `if (versionsCache[lawId])` guard forever.
+  if (list.length > 0) versionsCache[lawId] = list;
   return list;
 }
 
--- a/frontend/src/components/ScenarioBuilder.vue
+++ b/frontend/src/components/ScenarioBuilder.vue
@@ -2,8 +2,8 @@
 import { ref, computed, watch, nextTick, onBeforeUnmount } from 'vue';
 import { useDependencies } from '../composables/useDependencies.js';
 import { useScenarios, isScenarioMismatch } from '../composables/useScenarios.js';
-import { lawUrl } from '../composables/corpusUrls.js';
-import { apiFetchText } from '../lib/apiFetch.js';
+import { lawVersionsUrl } from '../composables/corpusUrls.js';
+import { apiFetchJson } from '../lib/apiFetch.js';
 import { useLatest } from '../lib/useLatest.js';
 import { parseFeature } from '../gherkin/parser.js';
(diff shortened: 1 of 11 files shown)


tdjager:
Re: *scope not recorded when /versions returns an empty array* (useEngine.js) — keeping this retryable is intentional, not an oversight. A missing dependency is typically being harvested in the background (`loadAllDependencies` fires `requestHarvestBatch` for it), and re-fetching `/versions` on the next `runDependencyLoad` is exactly how the newly-harvested law gets picked up — a negative-result/TTL cache would *delay* that pickup until it expired. The critical path already avoids redundant calls within a run: `ScenarioBuilder`'s `fetchLawVersions` caches the `[]` array in `versionsCache` (an empty array is truthy in JS, so the cache short-circuits the next call). So the only repeats are on `loadDependency`'s error path for a genuinely-missing law that hasn't been harvested yet — which is the case we *want* to keep polling. Leaving as-is.

PR #858, opened 2026-06-25
refactor: consume regelrecht-law-model in pipeline and harvester

tdjager:
## What & why

Advies 11 from the codebase audit: the canonical law-YAML model lives in the leaf crate `regelrecht-law-model` (already consumed by `engine`, `corpus`, `tui`). The two remaining consumers kept a partial mirror, so a schema change had to be hand-synced in several places. This routes the **executable** model through the single source while respecting the formatting/projection code that a naive parse→serialize roundtrip would break.

## pipeline — type the read/count side

**Before:** `count_article_stats` walked the law as untyped `serde_yaml_ng::Value` — `map.get("articles")`, `article_map.contains_key("machine_readable")` (stringly-typed field access).

**After:** it parses into `regelrecht_law_model::ArticleBasedLaw` and counts `articles.len()` / `articles.iter().filter(|a| a.machine_readable.is_some())`.

- The surgical **write** path is intentionally left untouched: production enrichment is done by an external LLM subprocess that edits the file directly, so there is no Rust-side model write to unify.
- Behavior change: a structurally-invalid law now surfaces as a parse error in the count instead of being silently undercounted — desirable, as this only runs on real harvested/enriched corpus files. Test fixtures updated to realistic minimal laws + an empty-`machine_readable` edge case.

## harvester — drop a duplicate, keep the write-projection

The writer is a deliberate write-projection: it emits `preamble`, `organisation`, `cvdr_id` and per-article `references` (which the executable `ArticleBasedLaw` does not carry) with bespoke quoting/indent post-processing, so it is **not** routed through `ArticleBasedLaw` (that would change byte output and lose data).

- **Before:** an internal `YamlReference` struct duplicated `types::Reference` field-for-field. **After:** it is removed and `types::Reference` is serialized directly (byte-identical — same fields, order, `skip_serializing_if`).
- Added `regelrecht-law-model` as a **dev-dependency** and two regression tests the repo lacked: golden **byte-identity** for a BWB and a CVDR law (previously only substrings were asserted), and a **conformance** test that the harvested output deserializes cleanly into `ArticleBasedLaw`.

## Out of scope

No new fields / reordering / `skip_serializing_if` churn on `ArticleBasedLaw` (it is deserialize-only today and harvested output must stay byte-identical). `harvester/src/types.rs` `Law`/`Article`/`Preamble`/`Reference` stay — they model the scraped *source* document, not the YAML document.

## Verification

`cargo fmt --check`, `clippy --all-features -Dwarnings`, `cargo check`, and the harvester + pipeline test suites are green. The new harvester golden test is the byte-identity regression proof.


tdjager:
Thanks for the review. Re: the `machine_readable: null` semantics note — the behavior (`{}` counts as enriched, `null` does not) is documented on `count_article_stats` and pinned by `test_count_article_stats_empty_vs_null_machine_readable` (commit 3e6d600, now rebased into 4232d37), and no corpus file uses the bare/null form today.

The suggested corpus-level guard against a hypothetical future `machine_readable: null` is intentionally left out of scope here: this PR is a model-dedup refactor, and a runtime/CI guard on LLM enrichment output belongs in the enrich-output validation path (or schema), not in this change. Noting it as a sensible follow-up rather than folding it in. No code change for this finding.

PR #859, opened 2026-06-25
feat(lawmaking): migrate UI to the NLDD design system

tdjager:
## What & why

`frontend-lawmaking` had drifted off the MinBZK NLDD design system. Its `App.vue`
wrapped everything in a **non-existent `<rr-page>` element** (a leftover from a
prefix-rename) and used **0** `nldd-*` components — all chrome was hand-rolled.
This is deliverable **A** of codebase-audit advies 15: put the standard UI back on
the design system.

## Changes

- **Page shell**: `<rr-page>` → `nldd-app-view` > `nldd-page` (sticky header).
- **Toolbar** (`nldd-toolbar`): playback navigation as `nldd-icon-button`s
  (`chevron-double-left` / `chevron-left` / `chevron-right` / `chevron-double-right`),
  a play/pause `nldd-button` (controlled by `isPlaying`, filled `primary` while
  playing), a step `nldd-tag`, zoom `nldd-icon-button`s (`plus` / `minus` /
  `arrow-2-counter-clockwise`), and an `nldd-segmented-control` for the view mode.
- **Detail panel**: `StageDetail` rewritten from a hand-rolled fixed panel into an
  `nldd-sheet` (right placement), driven declaratively from the selected stage. The
  last stage stays rendered through the close animation so the panel never flashes
  empty.
- **vite**: `isCustomElement` prefix `rr-` → `nldd-`.
- **Version**: `@nldd/design-system` `^0.8.62` → `^0.8.63` (aligned with `frontend`).

Net **−172 lines** (hand-rolled chrome CSS removed).

## Documented design-system exception

The **SVG GitFlow canvas** (`FlowDiagram.vue` / `FlowNode.vue` / `FlowConnection.vue` /
`FlowLegend.vue`, ~540 lines) is **kept hand-rolled** — the design system has no
graph-canvas equivalent. These files are untouched; they continue to own the
branch-colour tokens in `css/variables.css`.

Note: the design system has **no media-transport (play/pause) glyph**, so play/pause is
a labelled button rather than an icon button — a design adaptation, not a custom-CSS
shortcut.

## Remaining custom CSS (reported per CLAUDE.md)

All that is left on top of the design system, in the two migrated files:

| File | Rule | Why |
|------|------|-----|
| `App.vue` | `.app-logo` (32px height) | sizing the Rijkswapen mark — the toolbar has no logo slot |
| `App.vue` | `.diagram-area` (flex column, centre) | centres the SVG-canvas exception in the full-bleed section |
| `StageDetail.vue` | `.stage-detail__meta` (flex row, gap) | lays out the tag + subtitle row |
| `StageDetail.vue` | `.stage-detail__subtitle` (`margin:0`) | resets the rich-text paragraph margin |

(Pre-existing canvas CSS in the four Flow* components + `css/variables.css`/`main.css`
is part of the documented exception and is unchanged.)

Also a minor visual simplification: the per-branch coloured type badge is now a plain
`nldd-tag` (monochrome).

## Verification

- `npm run build` ✅
- Smoke-tested the built site: playback (play/pause/step/begin/end), zoom, view toggle
  (Eenvoudig/Uitgebreid/Wet open overheid), and the detail sheet all work; no console
  errors.

### Screenshots

Toolbar (single row, all controls):

toolbar

Detail sheet (node click):

sheet

Uitgebreid view:

advanced

### DIFF
--- a/frontend-lawmaking/src/components/StageDetail.vue
+++ b/frontend-lawmaking/src/components/StageDetail.vue
@@ -47,7 +47,7 @@
 </template>
 
 <script setup>
-import { ref, watch } from 'vue';
+import { ref, watch, nextTick } from 'vue';
 
 const props = defineProps({
   stage: { type: Object, default: null },
@@ -66,13 +66,18 @@ const shownStage = ref(null);
 // parent stays declarative (`:stage` / `@close`).
 watch(
   () => props.stage,
-  (stage) => {
-    if (stage) {
-      shownStage.value = stage;
-      sheetEl.value?.show();
-    } else {
+  async (stage) => {
+    if (!stage) {
       sheetEl.value?.hide();
+      return;
     }
+    shownStage.value = stage;
+    // Render the v-if="shownStage" content before opening so the sheet never
+    // animates in empty. `{ flush: 'post' }` wouldn't suffice here: this watcher
+    // itself sets `shownStage`, so the content render is still pending when the
+    // callback runs — awaiting nextTick flushes it first.
+    await nextTick();
+    sheetEl.value?.show();
   },
 );
 
--- a/frontend-lawmaking/src/components/StageDetail.vue
+++ b/frontend-lawmaking/src/components/StageDetail.vue
@@ -1,195 +1,98 @@
 <template>
-  <transition name="slide">
-    <div v-if="stage" class="stage-detail">
-      <div class="stage-detail__header">
-        <h2 class="stage-detail__title">{{ stage.lawLabel }}</h2>
-        <button
-          class="stage-detail__close"
-          @click="$emit('close')"
-          aria-label="Sluiten"
-        >&#x2715;</button>
-      </div>
+  <Teleport to="body">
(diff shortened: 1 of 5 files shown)


frontend-lawmaking/src/components/StageDetail.vue
```diff
@@ -1,195 +1,99 @@
...
+  () => props.stage,
+  (stage) => {
+    if (stage) {
+      shownStage.value = stage;
+      sheetEl.value?.show();
+    } else {
+      sheetEl.value?.hide();
+    }
+  },
+);
```
tdjager, line 82: Fixed in b959f3b0, but with `await nextTick()` rather than `{ flush: 'post' }`. Post-flush wouldn't actually help here: this watcher itself sets `shownStage`, so the `v-if` content render is still pending when the callback runs (post-flush only flushes renders triggered *before* the watcher fires). Awaiting nextTick flushes the content into the DOM first, then `show()` opens with content present.

PR #860, opened 2026-06-25
refactor(frontend): extract shared auth/color-scheme/fetch package

tdjager:
## What & why

The auth / color-scheme / fetch logic was **duplicated and diverged** between the
editor (`frontend/`) and admin (`packages/admin/frontend-src/`): two `useAuth`,
two `useColorScheme`, and a centralised `apiFetch` (editor, #795) with no admin
equivalent. This is deliverable **B** of codebase-audit advies 15: reconcile them
into one source consumed by all three frontends.

The editor is treated as canonical; **admin conforms to it**.

## New package — `packages/frontend-shared` (`@regelrecht/frontend-shared`)

Pure-ESM Vue, **no `@nldd` dependency**, `vue` as a peer, consumed as **raw source**
(`exports` → `src/index.js`, no build step). Exports:

- `apiFetch` / `apiFetchJson` / `apiFetchText` / `ApiError` — the editor's
  centralised implementation, verbatim.
- `useAuth` / `ensureAuthReady` — the editor's canonical implementation.
- `useColorScheme(persistence?)` — **owns** `<html data-scheme>` application + theme
  validation, with a pluggable persistence adapter (`{ theme, setTheme }`).
  Default = a localStorage adapter; the editor injects its server-backed
  `useUserSettings`.

## Consumers

| App | Consumes | Notes |
|-----|----------|-------|
| editor | apiFetch + useAuth + useColorScheme | `apiFetch.js` / `useAuth.js` become 1-line re-export shims (≈25 call sites unchanged); `useColorScheme` wraps the shared one with the server-backed adapter; the `data-scheme` watchEffect moved out of `useUserSettings` into the shared composable |
| admin | apiFetch + useAuth + useColorScheme | conforms to the editor; `authedFetch` stays admin-local glue (admin has no global 401 guard) and sources its redirect from the shared `login` |
| lawmaking | useColorScheme only | static site, no auth/API |

## Package strategy — root npm workspace

A root `package.json` (workspaces) + root `.npmrc` + a **single root lockfile** replace
the three per-frontend lockfiles. Chosen over a `file:` dependency because all three
Docker builds already use the repo root as context, so a bare specifier
(`@regelrecht/frontend-shared`) gives **identical dev/Docker resolution** and one source
of truth. (Accepted cost: each builder stage's `npm ci` installs all workspaces'
dev-deps — builder-stage only; runtime images ship just `dist`/`static`.)

## Docker (the main risk) — verified

All three frontend Dockerfiles now build via the workspace: copy the root
manifest + lockfile + every workspace `package.json` + the shared source, then
`npm run build -w <name>`. Runtime dist paths updated (`/app/frontend/dist`,
`/app/packages/admin/static`, `/app/frontend-lawmaking/dist`); the editor's WASM copy
target → `frontend/public/wasm/pkg/`. `.dockerignore` negations keep every workspace
`package.json` in context.

**Verified locally:**
- `npm run build` ✅ for all three (lawmaking, admin, editor)
- editor `vitest` ✅ **402/402**
- `docker build` ✅ lawmaking (full image) + admin & editor `frontend-builder` targets
  (incl. the WASM compile); dist/static produced at the expected runtime paths

## Notes
- Also aligns admin `@nldd/design-system` `^0.8.62` → `^0.8.63`.
- Admin's color-scheme localStorage key changes (`rr-admin-theme` → shared `rr-theme`),
  so an existing admin user's stored theme resets to `auto` once after deploy.


tdjager:
Re: "Admin Dockerfile has no guard on the vite output path" — disputing this one. The premise that `COPY --from` of a missing directory "succeeds silently" isn't how Docker behaves: `COPY --from=builder /app/packages/admin/static /app/static` **fails the build** if the source path doesn't exist (`failed to compute cache key: ... not found`). So a changed/mismatched `outDir` already surfaces as a loud build error at the COPY step — the extra `test -d` guard would be redundant. Addressed the other four findings in the latest commit.

tdjager:
Thanks — all three are non-actionable, disputing rather than churning the diff:

1. **`authedFetch` comment** — the comment is accurate: the *module-level* `useAuth()` call was removed, so importing the module no longer eagerly triggers the `/auth/status` fetch (which is exactly what it claims). As the review itself notes, the lazy `useAuth().login()` is "defensief correct", and `ensureAuthReady()` is `readyPromise`-guarded so the 401-path call never issues a second fetch. No behavioral change to make.

2. **localStorage key `rr-admin-theme` → `rr-theme`** — already called out in the PR description as an accepted one-time UX reset (the review acknowledges this). No change intended.

3. **Docker cache invalidation across images on any workspace `package.json` change** — the deliberate, documented trade-off of the npm-workspace approach (also acknowledged in the review). Builder-stage only; runtime images are unaffected.

CI is green and the correctness section confirms the `effectScope`/`WeakSet`/`readyPromise`/editor-facade wiring is sound.

tdjager:
Thanks — addressed the test-coverage finding:

- **🟠 No tests for the shared package** — added a vitest (happy-dom) suite to `packages/frontend-shared` (**35 tests**, run in CI via `npm test -w packages/frontend-shared`, and the package is now in the `ci`/`admin` path filters so a change there triggers it):
  - `apiFetch`: ok/non-ok → `ApiError` (status/body/contentType), `allowStatuses`, static + function `errorMessage`, network-error passthrough, `apiFetchJson`/`apiFetchText`.
  - `useColorScheme`: `applyColorScheme` (light/dark set, `auto`/invalid remove), `createLocalStoragePersistence` (valid/invalid read, custom key, write+validate, localStorage read/write throwing), the `effectScope`/`WeakSet` watcher applying + reacting + dedup, readonly `colorScheme`, and the no-arg localStorage default.
  - `useAuth`: `login()` return_url construction (default/explicit/PointerEvent-guard), `logout()`, and `checkAuth`/`ensureAuthReady` populating state + the `readyPromise` single-fetch.
- **🟡 module-level singletons hinder testing** — the default-path test demonstrates the `vi.resetModules()` mitigation.
- **🟡 `VALID_THEMES` exported with no consumer** — dropped from the package's public `index.js` (kept internal to the module).
- **🟡 `login()` open-redirect** — left as documented: the JSDoc already says to pass a relative/same-origin path, and as noted the backend validating `return_url` is the contract. Happy to add a client-side guard if you'd prefer the defence-in-depth.

PR #511, opened 2026-04-07
feat(engine): foreach (rfc-016) + native data source resolution (rfc-017)

anneschuth:
## Summary

Implements RFC-016 (FOREACH collection operations) and RFC-017 (native data source metadata). See PR #510 for the RFCs.

This PR enables pure-Rust orchestration: the engine reads inline `source.{table,field,fields,select_on,source_type,service}` metadata from YAML laws and resolves data source inputs itself, eliminating the need for host-side pre-resolution glue.

### FOREACH operation (RFC-016)

- New `ActionOperation::Foreach` variant in `article.rs` with `over`, `as`, `where?`, `combine?`, `value` fields
- Implementation in `operations.rs` using the existing child-context infrastructure for scoping
- Uses `MAX_ARRAY_SIZE` and `MAX_RECURSION_DEPTH` bounds
- Full execution trace integration

### Native data source resolution (RFC-017)

- `Source` struct in `article.rs` extended with optional `table`, `field`, `fields`, `select_on`, `source_type`, `service` fields (all `#[serde(default)]`, parsed automatically by serde)
- `RecordSetDataSource` in `data_source.rs` with builder pattern:
  - `key_field` for fast single-key lookup
  - `select_on` for multi-criteria filtering
  - `aliases` for input-name → column-name mapping
  - `array_field` with optional projection for FOREACH iteration
- `DataSource::query_native` trait method (default impl) + concrete impl on `RecordSetDataSource`
- `DataSourceRegistry::resolve_native` and `find_source_by_name`
- `service.rs::resolve_inputs_with_service` branches on `source.table.is_some()`: builds `Vec<SelectOn>` from YAML criteria (resolves `$param` refs against runtime parameters AND already-resolved inputs), calls `data_registry.resolve_native(...)`. Falls back to legacy `DataSourceRegistry::resolve` for YAMLs without `source.table`.
- `IN`/`NOT_IN` with an array subject (e.g. from `$arr.field` dot-notation projection) now matches element-wise: returns true if any element of subject is in the candidate list
- fix: topological input resolution order within an article (added in commit ac4ac616). Inputs can reference each other through `$name` in `source.parameters`, `source.select_on[*].value`, and `temporal.reference`. The resolver now topo-sorts inputs before resolving, falling back to YAML order on cycles. Unlocks alcoholwet Rotterdam, kieswet, and ~8 other downstream scenarios where later-declared inputs are referenced by earlier ones. See RFC-017 "Resolution order" section.
[...]

PR #878, opened 2026-06-26
refactor(engine): unify input-resolution onto a single ResolutionContext

tdjager:
## What & why

Picks up **Advies 13b** (engine: unify the two input-resolution paths — source of the "cycle-crash" bug class).

The engine had **two duplicated input-resolution paths**, each with its own cycle detector:

- **Path 1** — `ArticleEngine::resolve_input_sources` + `resolve_internal_reference`: resolved same-law `source.output` references by recursively spinning up sub-engines, with its own `visited: HashSet<article_number>` and depth cap (`MAX_RESOLUTION_DEPTH`).
- **Path 2** — `LawExecutionService::resolve_inputs_with_service` threading a single `ResolutionContext` (data-source → cross-law → internal tiers), used by **every production caller** (`wasm`, `tui`, `bin/evaluate`).

When the service called into `ArticleEngine` for action execution, the engine **reset to a fresh visited-set and depth 0**, so a cross-law↔internal cycle could evade *both* detectors. That seam is the cycle-crash class.

## Change

- Delete `resolve_internal_reference`; reduce `resolve_input_sources` → `check_input_sources`, which only validates that external (`source.regulation`) inputs were pre-resolved and otherwise leaves inputs untouched (lenient).
- Drop the now-dead `visited`/`depth` plumbing and the engine-level depth check.
- `ArticleEngine` now executes a **single article from already-resolved parameters**; `LawExecutionService`'s `ResolutionContext` is the **single resolution authority** — one cycle-detection scheme, one depth counter spanning cross-law + internal hops.

**No production behavior change**: every production caller already goes through the service, which pre-resolves inputs into `combined_params` before the engine runs. The only behavioral change is for *standalone* `ArticleEngine` with multi-article internal refs (tests/benches only): an unresolved internal input now surfaces as `VariableNotFound` rather than being auto-resolved.

## Tests

Six engine-level tests that relied on standalone multi-article internal resolution are migrated to the service layer, where value (basic/chain/with-parameters), self/indirect/three-way cycle, diamond and deep-chain depth cases are all covered.

## Relationship to #511

The todo marked 13b as blocked by #511 (RFC-016/017). On inspection #511 does **not** unify these paths — it adds a data-source tier alongside them and leaves both resolvers intact. So 13b is **independent of #511**; landing it first *shrinks* #511's surface (one resolver to extend, not two) rather than colliding.

## Verification

- `just format` ✓
- `just lint` (full workspace, `-Dwarnings`) ✓
- `just test` — 382 unit tests ✓
- `just bdd` — 11 features / 76 scenarios / 587 steps ✓
- `cargo bench -p regelrecht-engine --no-run` ✓

### DIFF
--- a/docs/src/content/docs/components/engine.md
+++ b/docs/src/content/docs/components/engine.md
@@ -270,8 +270,7 @@ The engine enforces compile-time security limits to prevent DoS:
 | `MAX_LOADED_LAWS` | 100 | Prevent memory exhaustion |
 | `MAX_YAML_SIZE` | 1 MB | Prevent YAML bombs |
 | `MAX_ARRAY_SIZE` | 1,000 | Prevent large array DoS |
-| `MAX_RESOLUTION_DEPTH` | 50 | Internal reference nesting |
-| `MAX_CROSS_LAW_DEPTH` | 20 | Cross-law reference nesting |
+| `MAX_CROSS_LAW_DEPTH` | 20 | Cross-law and internal reference nesting |
 | `MAX_OPERATION_DEPTH` | 100 | Operation nesting |
 
 ## Execution Receipt
--- a/packages/engine/src/config.rs
+++ b/packages/engine/src/config.rs
@@ -37,13 +37,16 @@ pub const MAX_YAML_SIZE: usize = 1_000_000;
 /// 1000 elements is sufficient for any reasonable law structure.
 pub const MAX_ARRAY_SIZE: usize = 1_000;
 
-/// Maximum depth for reference resolution in the service layer.
+/// Maximum combined depth for reference resolution in the service layer.
 ///
-/// Governs both cross-law reference chains and same-law (internal)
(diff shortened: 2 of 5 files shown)


tdjager:
Re: depth-check ordering asymmetry (internal `enter()`-then-check vs cross-law check-then-`enter()`, service.rs:2081–2086) — addressing the Significant finding (missing internal output now errors `OutputNotFound`, matching the external path; commit + regression test pushed).

On the Minor depth-check asymmetry: this is pre-existing code that this PR does not modify — the internal depth guard at 2081–2086 and the cross-law guard in `evaluate_law_output_internal` both predate this change (this PR's service.rs diff only adds tests). The off-by-one in the effective limit (20 vs 21 hops) is real but unrelated to the unification, and altering the internal limit here would be an out-of-scope behavior change to untouched code. Leaving it for a separate, dedicated cleanup so this PR stays scoped to the resolution-path unification.

PR #884, opened 2026-06-30
feat(editor): mint GitHub App tokens for traject repos

tdjager:
## Why

Today, coupling a **private** repo to a traject in the editor requires an operator to register a per-repo `CORPUS_AUTH_{REF}_TOKEN` env var on the editor component. That doesn't scale: env-var sprawl, a manual ops step per traject, and a long-lived PAT per repo. We also don't want to store tokens in the database.

ZAD offers no keyvault/secret-manager service (only postgres/redis/minio/keycloak/storage), so this builds the scalable alternative on top of GitHub itself: a **GitHub App** that mints short-lived, repo-scoped installation tokens on demand. Nothing long-lived is stored anywhere — only the app's private key, set once.

## What

A **hybrid** design: a provider-keyed factory becomes the top tier of the existing token resolver, with the current static chain kept intact as the fallback.

- **`packages/corpus/src/github_app.rs`** (new) — `GitHubAppAuth`: signs an app JWT (RS256), looks up the installation on the repo's owner (`/orgs/{owner}/installation`, falling back to `/users/{owner}/installation` for personal accounts), mints an installation token scoped to the one repo with `contents:write` + `metadata:read`, and caches both the installation lookup and the minted token with a TTL.
- **`packages/corpus/src/auth.rs`** — new `AppTokenMinter` trait, a **`ProviderAuthRegistry` factory** (`from_env`) that routes a source to its provider's minter by `SourceType`, and an async `resolve_token_async`. Resolution order: provider minter → per-source env var → `corpus-auth.yaml` → legacy `CORPUS_GIT_TOKEN`. The writable-own traject source keeps its **strict** static fallback (no legacy-token leak to a user-chosen repo).
- Threaded the registry through the registry async index/favorites paths and the editor-api wiring (`state.rs`, `main.rs`, `traject_corpus.rs`, `trajects.rs`, `corpus_handlers.rs`). The create-traject preflight now accepts "install the GitHub App on the repo" as an alternative to the env var.

### Factory / extensibility

`ProviderAuthRegistry` is provider-keyed — **today only GitHub is implemented**; GitLab / Azure DevOps slot in later as a new field + match arm, with call sites unchanged. Non-GitHub and local sources, and any deployment that doesn't configure the app, fall through to the existing env/file chain — behavior is byte-identical to before.

### Commit attribution

Unchanged and already correct: `persist` sets the commit `author` to the editor user, so App-minted commits still show the person who made the edit.

## Configuration (operator, one-time)

1. Register a GitHub App (org → Developer settings → GitHub Apps): permissions **Contents: Read & write** + **Metadata: Read**, webhooks off, installable on *Any account* (needed for cross-org single-repo installs).
2. Install it (org-wide, or "Only select repositories"). External orgs can install on exactly one repo.
3. Set on the editor component: `GITHUB_APP_ID` and `GITHUB_APP_PRIVATE_KEY` — the PEM **or its base64** (`base64 -w0 key.pem`, a single line — preferred where the platform passes env vars as a newline-separated blob, e.g. ZAD) — or `GITHUB_APP_PRIVATE_KEY_PATH` to a mounted PEM.

Testable with a **personal account** + a personal private repo — no org required.

## Tests

- `github_app.rs`: non-GitHub source → `None`; mint + cache (second call hits cache, no extra requests); org→user fallback; not-installed → `None` (wiremock).
- `auth.rs`: default registry has no minter; provider minter wins over the static chain; minter declines → falls back to per-source env var.
- `just format`, `just lint` (clippy clean), all corpus + editor-api unit tests green.

## Authorization model — deliberate decision

A code review confirmed the security invariant (no legacy `CORPUS_GIT_TOKEN` leak to a user-chosen repo) and the hybrid fallback. One behavioural point was reviewed and **consciously accepted**:

**Repo targeting is bounded by the App's installation scope, not by an in-app entitlement check.** Verified: per-traject membership (`traject_members`, owner/contributor) gates access to an *already-created* traject; it does **not** gate creation or repo choice. `POST /api/trajects` is gated only by the `editor-writer` role, and `resolve_writable_target` validates the *operator token's* push access — not whether the acting user is entitled to that specific repo. So with an **org-wide** ("All repositories") install, any `editor-writer` could couple a traject to any repo in that org.

**Accepted for now**: installation scope *is* the authorization boundary (installing the App = granting it). Recommended deploy mode for tighter control is a **selective install** (the operator picks which repos the App can see), which preserves the "someone deliberately enabled this repo" property of today's per-repo env var.

**Planned follow-up — #885**: bound repo choice in-app (operator allowlist / specific create-role / org↔user entitlement) enforced in `resolve_writable_target` (`trajects.rs:748`), so org-wide installs stay safe without relying solely on GitHub-side install scope. This is inherently a trust seam between two authz systems (editor roles/membership vs. GitHub App install scope) — the follow-up moves the seam from "installation scope" to an in-app entitlement. Parked for now.

### Other follow-ups (not blockers)
- Index/enumeration path resolves the writable-own source non-strictly (pre-existing; unreachable in practice because the create-time strict check guarantees a token exists).
- `mint_token` assumes the ~1h GitHub TTL rather than parsing `expires_at`; no single-flight on a cold-cache mint race (redundant mint only).

### DIFF
--- a/packages/Cargo.lock
+++ b/packages/Cargo.lock
@@ -2412,6 +2412,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "jsonwebtoken"
+version = "9.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5a87cc7a48537badeae96744432de36f4be2b4a34a05a5ef32e9dd8a1c169dde"
+dependencies = [
+ "base64 0.22.1",
+ "js-sys",
+ "pem",
+ "ring",
+ "serde",
+ "serde_json",
+ "simple_asn1",
+]
+
+[[package]]
 name = "kasuari"
 version = "0.4.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3138,6 +3153,16 @@ name = "peg-runtime"
 version = "0.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9555b1514d2d99d78150d3c799d4c357a3e2c2a8062cd108e93a06d9057629c5"
+
+[[package]]
+name = "pem"
+version = "3.0.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1d30c53c26bc5b31a98cd02d20f25a7c8567146caf63ed593a9d87b2775291be"
+dependencies = [
+ "base64 0.22.1",
+ "serde_core",
+]
 
 [[package]]
 name = "pem-rfc7468"
@@ -3885,6 +3910,7 @@ dependencies = [
  "base64 0.22.1",
  "flate2",
  "jsonschema",
+ "jsonwebtoken",
  "regelrecht-engine",
  "regelrecht-law-model",
  "reqwest 0.13.2",
@@ -4768,6 +4794,18 @@ name = "simdutf8"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e3a9fe34e3e7a50316060351f37187a3f546bce95496156754b601a5fa71b76e"
+
+[[package]]
+name = "simple_asn1"
+version = "0.6.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0d585997b0ac10be3c5ee635f1bab02d512760d14b7c468801ac8a01d9ae5f1d"
+dependencies = [
+ "num-bigint",
+ "num-traits",
+ "thiserror 2.0.18",
+ "time",
+]
 
 [[package]]
 name = "siphasher"
--- a/packages/corpus/Cargo.toml
+++ b/packages/corpus/Cargo.toml
@@ -24,6 +24,7 @@ jsonschema = { version = "0.46", optional = true }
 base64 = { workspace = true, optional = true }
 flate2 = { version = "1", optional = true }
 tar = { version = "0.4", optional = true }
+jsonwebtoken = { version = "9", optional = true }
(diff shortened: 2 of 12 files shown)


tdjager:
Thanks for the review — addressed the 🔴 (missing \`providers\` field in `list_scenarios_test.rs`), the 🟠 (added a `tracing::warn!` on the 422 negative-cache so an excluded repo is distinguishable from a typo), and the 🟡 double-JWT (now signed once per cold mint and threaded through both helpers) in `132eb7ba`.

Disputing the remaining two:

**🟡 Mutex on read-heavy caches** — kept as `tokio::sync::Mutex`. The critical sections are a single synchronous `HashMap` get/insert with **no `.await` held inside the guard**, and token resolution runs on the cold traject build/refresh path (backends are cached in `CorpusState`), not a per-request hot path. So reader serialization is sub-microsecond and effectively uncontended; `RwLock` would add overhead (and write-starvation risk) for no measurable gain. `Mutex` is the appropriate primitive for critical sections this short.

**🟡 Hollow `Source` probe in `trajects.rs`** — the `Source` is not hollow: `owner`/`repo` feed the App tier and `auth_ref` feeds the strict static fallback, and `resolve_token_async` needs all three to run the full hybrid (App → strict env/file) that the preflight requires. A narrower `token_for_github(owner, repo)` would cover only the App tier, forcing the handler to re-implement the strict fallback. And a new required field on `Source`/`GitHubSource` surfaces as a **compile error** at this construction site (Rust requires all fields), not a silent wrong-default; the filler fields (`branch`/`scopes`/`priority`) are unused by the resolution path.

Issue #885, opened 2026-06-30
Bound traject repo targeting: allowlist / create-role / org-user entitlement


tdjager:
## Context

PR #884 introduces GitHub App auth so a private repo can be coupled to a traject without a per-repo `CORPUS_AUTH_*` env var. That PR consciously accepts a widening of the authorization model, to be bounded by this follow-up.

**Verified today (origin/main):**
- Per-traject membership (`traject_members`, owner/contributor) gates access to an *already-created* traject only — it is checked on `get`/`update`/`delete` and per-request on the corpus read/write handlers (`trajects.rs:550–572`).
- It does **not** gate creation or repo choice. `POST /api/trajects` is gated only by the `editor-writer` role (`main.rs:372`), which most editor users have.
- `resolve_writable_target` (`trajects.rs:748`) only validates owner/repo *syntactically* and checks that the **operator token** can push — never that the **acting user** is entitled to that specific repo.

**Consequence:** with an org-wide ("All repositories") GitHub App install, any `editor-writer` can create a traject pointing at any repo in that org and get a `contents:write` token. Until this issue lands, the only real bound is GitHub-side install scope (use a **selective install**).

## Goal

Bound which repos a user may couple a traject to, in-app, so an org-wide install stays safe.

## Options to weigh (not mutually exclusive)

1. **Repo allowlist (select list).** Operator-maintained list of couplable `owner/repo` (config or DB table); reject anything else in `resolve_writable_target`. Simplest; bounds repo choice directly.
2. **Specific role for traject creation.** Gate `POST /api/trajects` (or just the repo-targeting path) behind a higher role (e.g. `editor-admin`) so only operators choose repos; ordinary `editor-writer`s work within existing trajects.
3. **Organization ↔ user entitlement mapping.** Model which orgs/users may target which repos — e.g. a user is entitled to repos under orgs they belong to, or an explicit org→allowed-users / org→allowed-repos mapping. Richer (supports federated/multi-org trajecten) but more design + data model.

Likely a combination: an **org/repo allowlist** (1/3) as the hard bound, optionally with a **create-role** (2) narrowing who can manage that list and create trajecten.

## Control point

`resolve_writable_target` in `packages/editor-api/src/trajects.rs:748` (creation-time repo entitlement) and/or the route role on `POST /api/trajects` (`main.rs`).

## Decision needed

[...]

PR #887, opened 2026-07-01
feat(editor): spike per-user GitHub OAuth for traject writes

tdjager:
## Why

PR **#884** gives the editor a **GitHub App** that mints all-access installation tokens, which means the editor itself must police *which user may write which repo* (the gap tracked in **#885**). This spike explores the alternative the team asked about: authenticate to GitHub **as the acting user** via OAuth. Then the editor never holds an all-access credential — GitHub enforces access, because a user's token can only touch what the user can already touch. The entitlement question disappears for the write path.

This is an **exploratory spike, separate from #884** ("let's see how it works and which approach we'd take"), scoped to where authorization actually matters: the **writable-own traject save + create-preflight path**.

## What

- **GitHub OAuth App link flow** (`packages/editor-api/src/github_oauth.rs`): `GET /auth/github/login` → `callback` → `status`, `POST /auth/github/disconnect`. Authorization Code flow, single-use CSRF `state` (constant-time compare), `repo read:org` scopes. Hand-rolled over `reqwest` (two calls) rather than wiring the typed `oauth2` client for a non-OIDC provider.
- **No token at rest — it lives in the browser.** The editor persists **no** GitHub credential server-side (no DB table, no session-store entry — the tower-sessions store is Postgres-backed, so "in the session" would still mean "in the database"). The callback seals the token with **ChaCha20-Poly1305** under `GITHUB_TOKEN_ENC_KEY` (`crypto.rs`) into an **HttpOnly, session-scoped browser cookie**, bound to the editor account that linked it: the browser carries an opaque blob it can neither read nor forge, and a leaked cookie value is useless without the server key. Disconnect = best-effort revoke at GitHub + clear the cookie.
- **Write-path threading**: every write request carries the cookie; the server opens it transiently and `WriteContext.token_override` supersedes the backend's baked token for that single write, so the commit authenticates *as the user* (`corpus/src/github_api_backend.rs`). Save handlers + the create-traject preflight (`trajects.rs::resolve_writable_target`) resolve the acting user's token from the request; the strict operator-token fallback that prevents `CORPUS_GIT_TOKEN` leaking to a user-chosen repo is preserved.
- **Relay mode** (`GET /auth/github/relay`, public): one registered OAuth-App callback serves production + every `prN` preview; the relay 302-forwards GitHub's response to the originating deployment, validated against a host-suffix allowlist so it can't become an open redirect.
- **Frontend**: `useGithubAuth` composable in `@regelrecht/frontend-shared` + a Connect/Disconnect GitHub item in the account menu (NDD components), gated behind the `github.user_oauth` feature flag.

### Behaviour switch — `GITHUB_USER_TOKEN_REQUIRED`

- **off (default)** — every write uses the backend's configured token, **byte-identical to today for every user, linked or not**. (This deliberately keeps a *linked* user's writes to the operator-managed central repo on the operator token, so they can't start 403-ing because the user lacks direct push access there.) Linking is offered + shown in the UI but inert for writes.
- **on** — "editor is not in the middle": every traject write uses the acting user's token; an unlinked/expired user is refused with **428 Precondition Required** (the frontend turns this into a "connect GitHub" CTA), never a silent fallback. A missing/tampered/foreign-account cookie all fail closed into the same 428.

## Config (operator, one-time)

1. Register a **GitHub OAuth App** (client id/secret; callback `…/auth/github/callback`, or the relay for multi-deployment setups).
2. Set on the editor component: `GITHUB_OAUTH_CLIENT_ID`, `GITHUB_OAUTH_CLIENT_SECRET`, `GITHUB_TOKEN_ENC_KEY` (base64 32-byte key, used to seal the cookie). Optional: `GITHUB_OAUTH_SCOPES`, `GITHUB_USER_TOKEN_REQUIRED=true`, `GITHUB_OAUTH_CALLBACK_BASE` + `GITHUB_OAUTH_ALLOWED_ORIGIN_SUFFIXES` (relay mode).

Testable with a **personal account + a personal private repo** — no org required.

## Design decision (deferred on purpose)

Recommended **classic OAuth App** because it best expresses "no all-access middleman" (the client secret grants nothing by itself; there is no installed app with standing org access). The alternative — reuse #884's App to mint *user-to-server* tokens — adds per-repo scoping + token expiry but reintroduces the installed app. Final call is the point of the spike.

## Out of scope (spike)

- App-fallback for users without GitHub (production would need it → complement, not replace #884).
- Migrating the **read** path to per-user tokens (reads stay on the existing resolver).
- The git-CLI/session-PR askpass path (writable-own uses the REST Contents API).
- Refresh-token rotation (classic `repo` tokens are long-lived; a refresh flow would live in the cookie alongside the access token).

## Tradeoffs

- **No credential at rest, anywhere** — the win, twice over: the editor holds no all-access token (vs #884), and it stores no per-user token either (vs the earlier iteration of this spike, which kept encrypted tokens in a `github_user_tokens` table). There is no token database to leak; a DB dump contains nothing.
- **Per-browser linking**: the token lives in that browser's session cookie — link again per device, and the link is gone when the browser session ends. Disconnect still revokes at GitHub.
- **Request-scoped only**: the server sees the token only while handling a request. A background job can never write as the user — acceptable for the spike scope (all writes are in-request), but a structural constraint to keep in mind.
- **Coarse scope**: classic OAuth `repo` is all-or-nothing across the user's repos (App user-tokens would be per-repo).
- **Adoption gate**: only works for users with GitHub write access.

## Tests

- `crypto.rs`: encrypt/decrypt round-trip, random-nonce, tamper + wrong-key rejection, key-length guards.
- `github_oauth.rs`: sealed-cookie round-trip (opaque value, account binding, garbage/absence rejection, expiry), cookie attributes (HttpOnly, session-scoped, Secure toggle for http-localhost dev), token exchange + provider-error + `/user` fetch (wiremock), relay origin-allowlist.
- `frontend-shared/useGithubAuth.test.js`: connect URL, status fetch/singleton, disconnect (8 tests; all 42 shared tests green).
- `just format`, `just lint` (clippy `--all-features`, clean), editor-api unit tests green, editor frontend builds.

## Notes for review

- `packages/editor-api/tests/save_annotations_test.rs` was already broken on `main` (stale `SESSION_KEY_ACTIVE_TRAJECT` import; editor-api integration tests aren't compiled in CI) and stays broken; `traject_reads_test`, `trajects_test` and `user_notes_test` compile and are updated for the new signatures.
- The earlier revision of this branch stored tokens encrypted in Postgres (migration `0025_github_user_tokens.sql`); that migration is **removed**. If the `pr887` preview DB already applied it, the migrator will refuse the missing version — delete the preview deployment for a fresh DB (known procedure).
- Not yet smoke-tested on the preview deployment — will do once `prN` is up.


tdjager:
### Preview smoke — no-regression / graceful-degradation path ✅

On the `pr887` preview (`editor-pr887-regel-k4c…`), logged in as the shared test writer:

- `GET /auth/github/status` → **200** `{connected:false, configured:false, expired:false, required:false}` — the new endpoint works, is auth-gated, and correctly reports "not configured" (no OAuth App env set on the preview).
- `GET /auth/github/login` & `POST /auth/github/disconnect` → **501** when unconfigured.
- App booted cleanly with migration `0022` applied; existing auth/session/library unaffected; the account menu hides the GitHub item when `configured:false`.

This exercises the default (feature-off) path end-to-end. **The full OAuth flow smoke is an operator step**: it needs a registered GitHub OAuth App plus `GITHUB_OAUTH_CLIENT_ID` / `GITHUB_OAUTH_CLIENT_SECRET` / `GITHUB_TOKEN_ENC_KEY` (and `GITHUB_USER_TOKEN_REQUIRED=true` to see the "editor not in the middle" mode) set on the `pr887` editor component — I can't register an org OAuth App or set secrets. Happy to walk through it once those are provisioned.

tdjager:
### Added: relay mode — one OAuth App for every preview + production

Follow-up on the "do I need an OAuth App per PR?" question. A classic OAuth App has **one** registered callback host, and GitHub matches it on **exact host** (subdomains don't inherit), so per-PR preview hosts would otherwise each need their own App. Rather than push people toward a GitHub App just for that, this adds a **callback relay** so a single App covers everything:

```
preview /login  ──redirect_uri = {CALLBACK_BASE}/auth/github/relay──►  GitHub
GitHub          ──(validates only the fixed relay host)────────────►  {CALLBACK_BASE}/auth/github/relay
relay (public)  ──302, origin from state, allowlist-checked────────►  {preview}/auth/github/callback
```

- `state` now carries `base64url({csrf, origin})`; the relay reads `origin`, checks it against `GITHUB_OAUTH_ALLOWED_ORIGIN_SUFFIXES`, and forwards `code`+`state`. The origin does the code exchange (it holds the secret) and the usual CSRF check.
- The relay is the **only** public GitHub route; it fails closed on a non-allowlisted/non-https origin, so it can't become an open redirect for the OAuth `code`.
- **Non-breaking**: unset `GITHUB_OAUTH_CALLBACK_BASE` → self-callback exactly as before. `from_env` refuses to start if a callback base is set without an allowlist.

**Extra config (only for relay mode):**

```
GITHUB_OAUTH_CALLBACK_BASE=https://editor.regelrecht.rijks.app
GITHUB_OAUTH_ALLOWED_ORIGIN_SUFFIXES=rijksapps.nl,editor.regelrecht.rijks.app
```

Register the App's single callback as `https://editor.regelrecht.rijks.app/auth/github/relay`. Then previews on `*.rijksapps.nl` and production all work under that one App. Tests added for the state round-trip, the self-vs-relay redirect_uri, and the origin allowlist (incl. `notrijksapps.nl` rejection + http rejection).

The GitHub-App variant (up to 10 callback URLs, no relay) remains the alternative — this just removes the "per-PR App" friction as an argument against the classic OAuth App.

tdjager:
Rebased op main en beide review-bevindingen verwerkt in `5e6789c9`:
- **user_write_token** faalt nu closed (503) bij ontbrekende pool in `require_user_token`-modus i.p.v. stil op het service-token terug te vallen.
- **with_marker** zet `?github=…` nu vóór een `#fragment`; test voor het fragment-geval toegevoegd.

Migratie hernummerd 0024→0025 wegens collisie met #913's `0024_user_notes`.

tdjager:
Beide review-bevindingen verwerkt in b909df1a: document-upload wordt nu fail-closed geweigerd (403) wanneer user-token-afdwinging aanstaat — bewust geen 428, want de background-conversie kan nooit namens de gebruiker schrijven, dus koppelen zou niets veranderen. `WriteContext::new` is aangesloten op de call site in `document_convert.rs`.

PR #894, opened 2026-07-02
feat(admin): harvest a law's model-dependency closure

tdjager:
## Why

Regelingen reached **only** through the machine-readable model — a `source.regulation` binding, an `implements`/`open_terms` (IoC) link, or a `legal_basis` anchor — are never auto-harvested. The recursive harvester follows **only `<extref>` BWB hyperlinks in the source text**. So a delegated regeling like **`regeling_standaardpremie`** (`BWBR0050536`), which fills `wet_op_de_zorgtoeslag`'s `standaardpremie` open term via an open delegation ("bij ministeriële regeling …"), never enters the corpus — and the zorgtoeslag chain can't resolve it at execution.

## What

New admin endpoint **`POST /api/harvest-dependencies`** (role `harvester-writer`): given a root law, walk its model-reference closure over the curated corpus and enqueue a harvest for every referenced law not already harvested. The existing extref recursion then continues from there.

- **corpus**: `collect_law_references()` (union of `implements` + `legal_basis` + `source.regulation`, the last via a recursive value walk since operation trees nest) and `extract_bwb_id()`. The reverse index reuses the precomputed `LoadedLaw.implements` (no per-request re-parse).
- **pipeline**: `find_bwb_id_by_slug` made `pub` for cross-crate reuse.
- **admin**: `harvest_deps` module — the closure, resolver, enqueue, handler + route.

### The closure follows edges in BOTH directions
- **Forward** — the law's own references (upstream inputs).
- **Reverse `implements`** — the lower regulations that implement this law's open terms. **Essential**: the higher law does *not* reference the delegated regeling forward (it only declares the open term); the link lives on the regeling. Without the reverse edge, `standaardpremie` — the whole point — would be missed.

Resolution `$id` → BWB: curated YAML `bwb_id` → `law_entries` slug lookup → `unresolved`. Enqueue goes through the canonical `request_harvest` (all dedup/exhausted/bookkeeping stays there). Returns a per-reference report `[{ id, resolved_bwb_id, source, action }]`.

## Scope / follow-ups
- **Phase 1 (this PR): on-demand.** Phase 2 (make it automatic in the harvest follow-up) is a deliberate follow-up.
- **SRU name-search fallback** (for referenced laws with no curated YAML) is *not* wired — such refs report `unresolved`. Left as a follow-up (`search_bwb` is an axum handler, not a plain fn; a match would only ever be `needs_confirmation`).
- **Yearly regelingen**: each year is a distinct BWB (2025 = BWBR0050536); resolution uses whichever curated version the corpus surfaces.
- `build_reverse_implements` is O(corpus) per request — fine for an on-demand admin action; a build-time index is the eventual optimization.

## Testing
- Unit: `collect_law_references` (all three ref kinds, zorgtoeslag-style legal_basis, dedup/self-exclusion, malformed tolerance), `extract_bwb_id`, reverse-implements discovery of the delegated regeling, and the status-classification gate. `cargo check`/`fmt`/`clippy` clean; corpus + admin lib tests green.
- End-to-end (to run on the preview): `POST /api/harvest-dependencies { "law_id": "wet_op_de_zorgtoeslag" }` → confirm it enqueues a harvest for `BWBR0050536`, it harvests, and a zorgtoeslag scenario using `standaardpremie` resolves instead of erroring "Law not found".

### DIFF
--- a/packages/admin/src/harvest_deps.rs
+++ b/packages/admin/src/harvest_deps.rs
@@ -1,0 +1,407 @@
+//! On-demand harvest of a root law's model-dependency closure.
+//!
+//! The recursive harvester only follows `<extref>` BWB hyperlinks found in a
+//! law's source text. Regelingen reached purely through the machine-readable
+//! model — a `source.regulation` binding, an `implements` IoC link, or a
+//! `legal_basis` anchor — are therefore never auto-harvested. A prime example
+//! is `regeling_standaardpremie` (BWBR0050536): `wet_op_de_zorgtoeslag`
+//! delegates its `standaardpremie` open term to it via `implements`/`open_terms`
+//! and nothing in the source text hyperlinks to it.
+//!
+//! This module walks that model-reference graph over the curated corpus,
+//! resolves each referenced `$id` to a BWB id, and enqueues a harvest for any
+//! law that has not already been harvested — reusing the canonical
+//! [`request_harvest`] entry point so all dedup/exhausted/bookkeeping semantics
+//! stay in one place.
+
+use axum::extract::State;
+use axum::Json;
+use regelrecht_corpus::source_map::{collect_law_references, extract_bwb_id};
+use regelrecht_pipeline::api::harvest::find_bwb_id_by_slug;
+use regelrecht_pipeline::harvest_request::{
+    request_harvest, HarvestRequestOptions, HarvestRequestOutcome,
+};
+use regelrecht_pipeline::law_status;
+use regelrecht_pipeline::{LawStatusValue, PipelineError, Priority};
+use serde::{Deserialize, Serialize};
+use std::collections::{HashMap, HashSet, VecDeque};
+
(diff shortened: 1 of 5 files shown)


tdjager:
Parkeren/sluiten: verkeerde aanpak. De discovery van gedelegeerde regelingen (bijv. `regeling_standaardpremie`) hoort niet in de admin — die heeft in deployment geen corpus. Sterker: de koppeling `open_terms`/`implements` is een **product van enrichment** (de machine-readable interpretatie), dus de discovery hoort op **enrichment-moment** te gebeuren (de enrichworker heeft corpus + DB). Voor de directe behoefte is `BWBR0050536` inmiddels rechtstreeks via `POST /api/harvest-jobs` geharvest + geënricht. De corpus-helpers (`collect_law_references`, reverse-implements) blijven herbruikbaar mocht de enrichment-side variant er komen.

PR #901, opened 2026-07-02
feat(pipeline): harvest related legislation discovered during enrichment

tdjager:
## Why

Regelingen reached **only** through the machine-readable model — a `source.regulation` binding, a `legal_basis` anchor, or an open delegation (`open_terms`/`implements`, e.g. `regeling_standaardpremie` filling `wet_op_de_zorgtoeslag`'s `standaardpremie`) — are never auto-harvested. The recursive harvester follows **only `<extref>` BWB hyperlinks in the source text**, and those model links don't exist in the text — they are a **product of enrichment**. So the discovery must happen where the model is created: the enrichworker (which has corpus + DB). (A prior admin-side closure endpoint, #894, was closed — the deployed admin has no corpus.)

## What

The enrichment agent now **returns the related legislation it needs** in a **result envelope** — a `.enrichment-result.yaml` sidecar the agent writes, kept deliberately **out of the law YAML** so the law artifact stays schema-conformant (extensible for future enrichment return-values). After a successful enrich, the worker:

1. reads the envelope's `related_legislation`,
2. resolves each entry to a BWB id — **hybrid**: agent `bwb_id` → `law_entries` slug lookup → single-hit SRU title search (`>1` → `needs_confirmation`, skip),
3. enqueues a follow-up harvest per resolved id via the canonical dedup path.

Newly harvested laws auto-enrich (existing `ENRICH_AUTO_ENQUEUE`) and return **their** related legislation → the dependency graph fills itself by recursion.

### Depth & priority
`EnrichPayload.depth` carries recursion depth `harvest → enrich → harvest`; each related harvest is `depth+1` at `priority = 40 − (depth+1)`, so deeper (more speculative) nesting yields to shallower/interactive harvests (`Priority::new` clamps `0..=100`; queue is `ORDER BY priority DESC`).

### Safety / guards
- **Opt-in**: `HARVEST_RELATED_LEGISLATION` (off by default) + `RELATED_HARVEST_MAX_DEPTH` (default 2).
- Reuses `ENRICH_DAILY_LIMIT` (spend) and `create_harvest_job_if_not_exists` (dedup); `harvest_exhausted` skipped.
- **Best-effort**: the whole block runs *after* the enrichment is committed and swallows every error into logs — it can never fail the enrichment or the worker loop.

### Also
- `search_bwb_by_name(client, q)` extracted from the axum handler (reused by both); `find_bwb_id_by_slug` made `pub`; slug hits re-validated as BWB (CVDR skipped, so no malformed harvest).
- **No law-schema change.** `.claude/skills/law-generate/SKILL.md` instructs writing the envelope.

## Known limitations
- **Shared depth counter** with the extref harvester → the loop fires for roots/shallow laws; a `related_depth` field is the clean follow-up.
- **Unambiguous-but-loose SRU match** accepted as-is (opt-in, harvest-only, lands on a reviewable `enrich/{provider}` branch).

## Testing
- **Unit** (52 pipeline tests green): envelope (de)serialization (full/missing/absent/malformed → empty), `is_valid_bwb_id`/`slugify`, `related_harvest_priority` (clamps, −1/level), depth round-trips.
- **End-to-end (to run on the preview)**: set `HARVEST_RELATED_LEGISLATION=true`, (re-)enrich `wet_op_de_zorgtoeslag` → confirm it writes `.enrichment-result.yaml`, a follow-up harvest for the delegated regeling is enqueued (depth 1, priority 39), harvested, then auto-enriched. Verify `ENRICH_DAILY_LIMIT` still bounds spend.


tdjager:
Both review findings addressed in 4a24029f: the CVDR slug hit now returns `Unresolved` instead of falling through to the SRU name search, and the harvest summary log separates `already_queued`/`exhausted` skips. Also dropped RFC-025 (this is an implementation detail, not an RFC).

PR #903, opened 2026-07-03
feat(pipeline): guard enrichment against a stale base branch

tdjager:
## Problem

Enrichment runs on shared `enrich/{provider}` branches (`enrich/claude`, `enrich/opencode`). Those branches accumulate enriched laws over time on top of whatever base they were born from — and nothing checked that the base was still current. In one incident `enrich/claude` was sitting on `main` while harvesting had moved to `development`: the branch was **7911 commits and a whole schema/article-granularity generation behind**. Every re-enrichment silently regenerated `machine_readable` on top of a stale, wrong-shaped base, quietly discarding possibly hand-validated work.

## The fix

Record provenance and refuse to enrich against a moved base:

- **Record `source_hash`** — the git blob SHA of the base-branch law YAML the enrichment was generated from — into each law's `.enrichment.yaml`.
- **Before re-enriching a law**, compare its current base blob SHA to the recorded `source_hash` via a pure `decide_base_action`:
  - **new law** (not yet on the enrich branch) → check it out fresh from the base;
  - **unchanged** (recorded hash == base blob SHA) → proceed, keep the existing enrichment;
  - **drifted / missing provenance** (hash differs or absent) → **fail the job loudly** as `enrich_failed`. Never auto-overwrite a possibly-validated enrichment.
- The base reference is the existing `pick_enrich_base` result — `development` today, `main` if the corpus is migrated later — so **no base-selection code changes**, the guard just observes the base that would already have been used.

The freshness check operates on the exact law file path (via `is_tracked` / `fetch_base_blob_sha` resolving a single blob), so a newly harvested version of an already-known law is judged on its own path rather than masked by a sibling version in the same directory.

## Out of scope / future work

- Enrichment validation-state tracking (knowing *which* enrichments a human has validated).
- A schema-migration path for laws enriched against an older schema generation.
- The `development` → `main` migration itself.
- Whole-branch drift monitoring (this guards per-law at enrich time, not the branch as a whole).

## Notes

The 9 commits are TDD-structured (pure decision function and error variant first, then corpus primitives, then the glue, then the loud-failure/retry wiring, then a final refactor). This PR adds one end-to-end test wiring `create_enrich_corpus` against a real bare-git remote whose `enrich/test` branch carries a stale `.enrichment.yaml`, asserting the guard returns `BaseDrift`.

Design + plan: /workspace/superpowers-specs/2026-07-02-enrich-base-freshness-guard-*.md (local reference)

### DIFF
--- a/packages/corpus/src/client.rs
+++ b/packages/corpus/src/client.rs
@@ -671,6 +671,36 @@ impl CorpusClient {
         Ok(())
     }
 
+    /// Fetch the base branch shallowly and return the git blob SHA of `path` on
+    /// it. Uses `FETCH_HEAD` (the single-branch enrich clone never creates
+    /// `origin/<base>`). Blob SHA is content-addressed, so it is comparable
+    /// across clones.
+    pub async fn fetch_base_blob_sha(&self, base_branch: &str, path: &str) -> Result<String> {
+        self.run_git(&["fetch", "--depth", "1", "origin", base_branch])
+            .await?;
+        let sha = self
+            .run_git_output(&["rev-parse", &format!("FETCH_HEAD:{path}")])
+            .await?;
+        Ok(sha.trim().to_string())
+    }
+
+    /// Whether `path` is tracked in the current checkout/branch.
+    pub async fn is_tracked(&self, path: &str) -> Result<bool> {
+        let listed = self.run_git_output(&["ls-files", "--", path]).await?;
+        Ok(!listed.trim().is_empty())
+    }
+
+    /// Check `path` out of the most recent `FETCH_HEAD` and unstage it, so it
+    /// does not pollute an empty `git status --porcelain` check before the
+    /// enrichment commits. Call `fetch_base_blob_sha` first so `FETCH_HEAD` is
+    /// set.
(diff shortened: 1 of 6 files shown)


tdjager:
Both review findings addressed in `ec06da70`:

- 🟠 **Mass-drift of pre-PR enrichments** — added a distinct `BaseAction::AdoptBaseline`. A tracked law with no usable recorded provenance (absent/empty `source_hash`, i.e. every enrichment from before this guard) is now *grandfathered*: it adopts the current base blob SHA as its baseline and proceeds like `Skip` (no fresh checkout, existing enrichment kept), instead of terminal-failing as `BaseDrift`. The baseline is stamped on the next `.enrichment.yaml` write, so real drift is detected from that point on. Only a *recorded* mismatch (`Some(h) != base`) still `Drift`s — the e2e test uses a non-empty stale hash, so it stays `BaseDrift`.
- 🟡 **'retryable' comment** — corrected to 'terminal failure (requires human re-enrich)'.

Unit test renamed `..._is_drift` → `..._adopts_baseline` and asserts `AdoptBaseline` for both `None` and `Some("")`.

PR #913, opened 2026-07-06
feat(editor): personal notes API with personal/public saving distinction

tdjager:
## Summary

API for private per-user notes (*persoonlijke notities*) on a law or regulation, plus a **unified notes endpoint**: the UI saves and reads notes through one channel and never needs to know where a note is stored. **API-only** — the editor UI will be built separately by someone else.

### The unified contract: one endpoint, visibility as data

`PUT/GET /api/trajects/{ref}/corpus/laws/{law_id}/annotations` is the single notes channel. Each note carries an optional `"regelrecht:visibility"` property (vendor-prefixed like the existing `regelrecht:hint`):

| `regelrecht:visibility` | Storage |
|---|---|
| `"personal"` | Postgres (`user_notes`), keyed by `accounts.id` — never leaves the database |
| `"public"` or absent | Stand-off sidecar YAML on the traject's git branch (RFC-018 Decision 1), append-only, verbatim base bytes |

- **PUT** partitions the submitted array per note; the marker is stripped before anything reaches the sidecar; an unknown visibility value is a 400 (a typo must never publish a note meant to be personal). Personal inserts are content-deduped so retries stay idempotent, mirroring the sidecar dedup. Response gains `personal_saved`.
- **GET** returns one YAML document: the sidecar (bytes verbatim) with the caller's personal notes appended, each marked `regelrecht:visibility: personal` (+ server `id`/`created`/`modified`). Works when no sidecar exists yet. The existing client-side resolver consumes it unchanged.
- **Safety property**: a personal-marked note can never end up in git — even a naive round-trip of the merged GET document back into a PUT routes it to the database (where dedup makes it a no-op).
- Item-level ops on personal notes (edit/delete an individual note, which the sidecar side has no analogue for) stay available at `GET/POST /api/user/notes/{law_id}` and `PUT/DELETE /api/user/notes/{law_id}/{note_id}`.

Both sides speak the same W3C Web Annotation model (RFC-005): `TextualBody` (`value` + `format`, markdown by default) and an optional `target.selector` (TextQuoteSelector, stored verbatim as JSONB) — anchored notes keep their anchoring on either side. Personal notes support single `TextualBody` commenting/questioning notes; linking/tagging/multi-body notes are public-only and get a concrete 400.

### Data model (follows RFC-005 / RFC-018)

Migration `0024_user_notes.sql`: `user_notes` table mirroring the annotation model (`motivation`, `body_value`, `body_format`, `selector` JSONB), FK to `accounts` (`ON DELETE CASCADE`), composite index `(account_id, law_id)`, `updated_at` trigger.

### Personal-store details

- Reads at `editor-reader`, mutations at `editor-writer`, identity via the existing `account_middleware`; all queries scoped to `account_id` (foreign note ids 404, no existence leak).
- Validation: 64 KiB body cap, 8 KiB selector cap (JSON object with a `type`), whitelisted formats (`text/markdown`, `text/plain`) and motivations (`commenting`, `questioning`), `law_id` slug allowlist `[a-z0-9_.-]`.
- 200-notes-per-law cap, enforced under a per-(account, law) advisory lock (race-free; 409 on overflow).
- `PUT` on the item route: `value` required; `format`/`motivation` absent = keep; `selector` absent = keep, explicit `null` = detach (double-`Option` wire contract, pinned by a serde test).

### Drive-by fix

The four existing editor-api test files no longer compiled on main (`AppState` gained `harvest_admin_url` without the tests following — they are not run in CI). They now set the field. `save_annotations_test.rs` has additional pre-existing API drift (`SESSION_KEY_ACTIVE_TRAJECT`, old `save_law` signature) that is out of scope here; follow-up needed.

## Test plan

- [x] `cargo test -p regelrecht-editor-api --test user_notes_test` — 13/13 green (fresh DB through migration 0024): W3C shape, privacy/IDOR, cap, selector roundtrip/detach, absent=keep, dedupe for the unified save, annotation→note mapping, visibility marker on merged values
- [x] Bin unit tests 95/95 green, incl. visibility partitioning (routing + marker stripping + unknown-value 400) and the author-spoofing signature assertion updated for the new handler signature
- [x] `just editor-api-fmt` / `editor-api-lint` / `editor-api-check` clean
- [x] Frontend suite still green after removing the UI (430 tests)
- [x] Smoke test of the unified PUT/GET on the pr913 preview: personal-marked PUT → `personal_saved: 1` + no git write; merged GET carries the note with the visibility marker (works with absent sidecar); global corpus read stays clean (404, nothing leaked to git); retried PUT dedupes (`personal_saved: 0`); cleaned up via the item route


packages/pipeline/migrations/0024_user_notes.sql
```diff
@@ -0,0 +1,31 @@
+-- Persoonlijke notities: private per-user notes on a law, stored server-side
+-- so they never leave the database (unlike traject annotations, which are
```
tdjager, line 2: Disputed: this follows the repo-wide convention — none of the 23 existing migrations use `IF NOT EXISTS` on tables/indexes/triggers (0001, 0013, 0014, 0023 are all bare `CREATE`; the only `IF NOT EXISTS` in the corpus is `CREATE EXTENSION`). Idempotency is provided by the sqlx-migrate runner (checksummed run-once under an advisory lock), and a bare `CREATE` failing loudly against a partially-initialised schema is preferable to `IF NOT EXISTS` silently no-op'ing over a divergent table. Also, `CREATE TRIGGER IF NOT EXISTS` does not exist in PostgreSQL, so the suggested fix can't be applied as written. If REVIEW.md's "idempotent" line should mean per-statement idempotency, that's a repo-wide decision (REVIEW.md vs all 24 migrations), not something for this PR.

tdjager:
Re: review findings on the `create` cap and `validate_law_id` — both addressed in 599ef133:

- **Cap atomicity**: correct — the single `INSERT..SELECT..WHERE count < cap` statement was not race-free under READ COMMITTED and the comment overclaimed. `create` now serializes per (account, law) with `pg_advisory_xact_lock` inside a transaction, so the 200-cap is strictly enforced.
- **`validate_law_id`**: now allowlists `[a-z0-9_.-]` (all corpus `$id`s are lowercase snake_case slugs; verified against origin/main), so junk can't reach storage or the echoed `regelrecht://` URI.

tdjager:
Re: round-2 findings — addressed in 14a528cc:

- **`selector: null` silently kept the anchoring**: correct. `NoteRequest.selector` is now a double `Option` (absent = keep, explicit `null` = detach; `CASE` instead of `COALESCE` in the UPDATE). The wire contract is pinned in a serde test and the detach path in an integration test.
- **`ORDER BY created_at` now says `ASC` explicitly.**
- **law_id sad paths** are now also exercised via `create` and `remove`.
- **CHECK constraints on `motivation`/`body_format`**: deliberately not added — as the review itself notes, all existing tables (trajects, user_settings) validate in the application layer only, and editor-api is the sole writer. Keeping 0024 unchanged also avoids another preview-DB reset for an in-flight migration edit.

tdjager:
Re: round-3 findings — both addressed in cf994bb0:

- **Partial personal-note commit**: personal notes in the unified save are now a two-phase atomic batch — the whole batch is validated before any store work, then inserted in one transaction (all-or-nothing; pinned by `insert_notes_batch_is_all_or_nothing`, covering both the invalid-note and the cap-crossing case).
- **`no_change` semantics**: documented on the `SaveResponse` field as a sidecar/git-only concept that can be `true` while `personal_saved > 0`; clients must gate on the combination. Kept the field name — renaming would break the existing frontend consumers of `no_change`.

PR #928, opened 2026-07-10
perf(frontend): dedupe law body and laws-list fetches on editor open

tdjager:
## What & why

Part of the editor↔GitHub performance track (follow-up to the instrumentation in #919). The editor-open waterfall showed two avoidable duplicate requests fired on every editor mount:

1. **Law body fetched twice**: `useLaw().load()` fetches the routed law while the persisted-tab label loader calls `fetchLaw()` for the same law id in the same tick — the shared cache is still empty at that point, so both hit the network.
2. **`laws?limit=1000` fetched twice**: `EditorView` kept a private `loadCorpusLaws()` fetch for the failed-law display name, next to the shared per-scope cache in `useCorpusLaws` that `MachineReadable` already uses.

## How

- **`useLaw.js`**: a module-level in-flight map (`pendingLawFetches`), shared by `load()` and `fetchLaw()`. Simultaneous callers share one GET; entries are dropped when the request settles so a failed fetch never pins a rejected promise. `load()` keeps its always-fresh semantics (it never *reads* the cache, it only shares the in-flight request), and the fresh entry is cached under both the requested id and the body's resolved `$id`.
- **`EditorView.vue`**: the private corpus-laws fetch is replaced by `useCorpusLaws(activeTrajectRef)` — same module-level per-scope cache as `MachineReadable`, so an editor mount fires one list fetch total. `failedLawName` now goes through the composable's `displayName` (`display_name || name || humanizeLawId`), which is a strict improvement over the previous `name || humanizeLawId`.

## Verification

- `npm run -w frontend test`: **440 passed (46 files)** ✓
- Live smoketest on the preview deployment follows once it's up (network tab should show exactly one law-body GET and one `laws?limit=1000` GET per editor mount).

### DIFF
--- a/frontend/src/EditorView.vue
+++ b/frontend/src/EditorView.vue
@@ -3,7 +3,7 @@ import { ref, computed, reactive, watch, watchEffect, nextTick, onBeforeUnmount 
 import { useRoute, useRouter, onBeforeRouteUpdate } from 'vue-router';
 import * as yaml from 'js-yaml';
 import { useLaw, fetchLaw } from './composables/useLaw.js';
-import { lawsListUrl } from './composables/corpusUrls.js';
+import { useCorpusLaws } from './composables/useCorpusLaws.js';
 import { useEngine } from './composables/useEngine.js';
 import { useAuth } from './composables/useAuth.js';
 import { useTrajects } from './composables/useTrajects.js';
@@ -20,7 +20,7 @@ import {
   clearEditorChrome,
 } from './composables/useAppChrome.js';
 import { SUPPORT_EMAIL } from './constants.js';
-import { apiFetch, apiFetchJson } from './lib/apiFetch.js';
+import { apiFetch } from './lib/apiFetch.js';
 import { humanizeLawId } from './lib/lawName.js';
 import { useLatest } from './lib/useLatest.js';
 import ArticleText from './components/ArticleText.vue';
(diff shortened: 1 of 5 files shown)


tdjager:
Smoketested on the pr928 preview: after the library→editor tab switch the waterfall now shows exactly **one** law-body GET and **one** `laws?limit=1000` GET (baseline on pr919 showed 2× each). Annotations/scenarios calls unchanged, as expected — those are addressed server-side in #929.

tdjager:
Smoketested on the pr928 preview: opening the editor fires no duplicate API calls (on pr929-preview without this change, the laws-list fetch still fires twice). Marking ready for review.

tdjager:
Fixed in 1fec688f: both dangling loadCorpusLaws() calls removed — the traject watch relies on useCorpusLaws' reactive re-scope; onSearchHarvestAvailable now awaits a new useCorpusLaws refresh() (cache bust) so the fresh law still gets its real display name. Stale comment updated, tests added (useCorpusLaws.test.js + dedup tests in useLaw.test.js).

tdjager:
Both review findings addressed in c9f74950: `load()` now snapshots the traject ref before its awaits, so the direct-URL cache write can no longer key on a trajectRef mutated by a concurrent `switchLaw` (regression test included, verified to fail on the pre-fix code); and `refresh()` is single-flighted per scope and serialized behind an in-flight fetch, so it can no longer orphan a pending request (duplicate concurrent GET) or let its late settle overwrite the refreshed list. Follow-up review round on c9f74950 reported no new findings.

PR #929, opened 2026-07-10
perf(editor): stale-while-revalidate refreshes and snapshot caches for sidecar reads

tdjager:
## What & why

Follow-up to the latency instrumentation in #919. Live `Server-Timing` measurements on the pr919 preview showed where the editor's GitHub-backed read path actually spends its time:

| pain point | measured |
|---|---|
| synchronous 60s-TTL index refresh | **1.7–1.9s**, paid by one unlucky request per minute per traject |
| synchronous changed-laws Compare on TTL expiry | 0.6–1.0s |
| annotations GET (uncached, incl. 404) | 250ms–**5.7s** per law open |
| scenario listing + per-file reads (uncached, serial) | 0.6s–**7.0s** per law open |

GitHub Contents API calls have a heavy latency tail (baseline ~250–600ms, spikes of 3–7s), so the fix is to keep them off the request path entirely.

## How

**Stale-while-revalidate for the index snapshot** — `get_or_build` no longer awaits the re-enumeration on TTL expiry: the first request past the deadline spawns a background refresh task (single-flighted by an `OwnedMutexGuard` on the slot's build lock) and serves the stale snapshot, like every concurrent request already did. A failed refresh re-arms the stale snapshot for another TTL window (unchanged degrade-to-stale stance, now unit-tested via the extracted `apply_refresh_outcome`). Only the very first build of a traject still blocks — there is nothing to serve yet.

**Same pattern for changed-laws** — `changed_law_ids` serves the cached diff immediately and recomputes in the background (AtomicBool single-flight with an RAII `ResetOnDrop` guard so a panicking task can't latch the flag). `save_law` now folds the saved law id into the cached diff (`record_changed_law`) instead of invalidating it, so the sidebar's next load no longer pays a synchronous Compare after every save.

**Per-snapshot caches for sidecar reads** — `TrajectCorpus` gains two bounded FIFO caches, per index snapshot (so upstream edits converge within one TTL window, exactly like law bodies):
- `sidecar_cache`: annotations sidecars and individual scenario files, **including negative (404) results** — "this law has no annotations" was a full GitHub round-trip per law open;
- `scenario_list_cache`: the per-law scenario listing (2× `list_files` + a read per file to rebuild).

Coherence: saves store the just-written body (read-your-writes), deletes drop the entry (the next read may legitimately fall back to a seed copy). A **write-generation counter** guards against the read/write race where a slow in-flight GET would re-cache pre-save bytes over a fresh save (which would 412-loop the saver's next If-Match until the snapshot swap): GET-path stores capture the generation before their backend read and are dropped if a write bumped it meanwhile. Write-path `If-Match` precondition reads deliberately bypass all caches.

**`gh_list` Server-Timing phase** for `list_files` — the listing round-trips were the uninstrumented gap between `gh_get` and `total` in the #919 measurements.

## Expected effect (to verify on the preview)

- Law open (warm traject): annotations + scenarios drop from ~3–8 serial GitHub round-trips to 0 within a snapshot window.
- The periodic ~1.8s stall on a random request per minute disappears.
- `changed-laws` after a save no longer costs a synchronous Compare.

## Verification

- `cargo test -p regelrecht-editor-api --lib`: **72 passed** (10 new tests: background-refresh swap, failure re-arm, changed-laws SWR + failure re-arm + fold semantics, generation guard, cache reset per snapshot, bounded-cache eviction hygiene) ✓
- `cargo clippy --all-features` with `-Dwarnings` (editor-api + corpus): clean ✓
- `cargo fmt --check` ✓
- An adversarial code review ran before push; the one major finding (the read/write cache race above) is fixed with the generation guard, plus three minor findings (RAII flag reset, bounded-cache remove/re-insert eviction edge, fold-undo race documented as bounded).
- Live before/after measurement on the preview deployment follows (same probe set as the #919 baseline).

## Known limitations

- GitHub answers 404 for a private repo when a token is invalid/expired, so during a token hiccup existing sidecars can be negative-cached for up to one snapshot window (≤ ~60s) per key. Accepted: same conflation existed uncached, and errors (403/5xx) are never cached.
- A changed-laws recompute whose Compare call was in flight during a save can briefly drop that save from the "Bewerkt in dit traject" list; self-heals within one TTL (documented in code).

### DIFF
--- a/packages/corpus/src/github_api_backend.rs
+++ b/packages/corpus/src/github_api_backend.rs
@@ -304,18 +304,23 @@ impl RepoBackend for GitHubApiBackend {
         Ok(())
     }
 
+    #[tracing::instrument(name = "gh_list_files", skip_all, fields(dir = %dir.display()))]
     async fn list_files(&self, dir: &Path, extension: Option<&str>) -> Result<Vec<FileEntry>> {
         let api_dir = self.api_path(dir)?;
         let mut inner = self.inner.lock().await;
-        let entries = inner
-            .fetcher
-            .list_directory(
+        // One Contents API directory GET — feeds the `gh_list` Server-
+        // Timing phase so listing round-trips (scenario listings) are
+        // visible next to `gh_get` instead of vanishing into `total`.
+        let entries = timing::measure(
+            "gh_list",
+            inner.fetcher.list_directory(
                 &self.full_repo(),
                 &self.branch,
                 &api_dir,
                 self.token.as_deref(),
-            )
-            .await?;
+            ),
+        )
+        .await?;
         let mut out: Vec<FileEntry> = entries
             .into_iter()
             .filter(|e| e.entry_type == "file")
--- a/packages/editor-api/src/corpus_handlers.rs
+++ b/packages/editor-api/src/corpus_handlers.rs
@@ -746,7 +746,14 @@ async fn list_scenarios_in_scope(
                 let target_law_ids =
                     match read_traject_scenario_cached(traject, law, &relative_path).await {
                         Ok(Some(content)) => extract_target_law_ids(&content),
-                        Ok(None) => Vec::new(),
+                        // Listed but unreadable on every routing leg: a
+                        // ghost. GitHub's directory listing is eventually
+                        // consistent, so right after a delete commit the
+                        // (stale) listing can still name the file while
+                        // its content GET already 404s. Skip it — keeping
+                        // it would serve a listing entry whose GET can
+                        // only 404.
+                        Ok(None) => continue,
                         Err((_, err)) => {
                             tracing::warn!(
                                 law_id = %law_id,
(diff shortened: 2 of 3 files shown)


tdjager:
## Live measurements on the pr929 preview (first deploy)

Same probe set as the #919 baseline, traject `zorgtoeslag private repo`, law `wet_op_de_zorgtoeslag`:

| probe | baseline (pr919) | pr929 |
|---|---|---|
| annotations GET (404 case) | 0.25–**5.7s**, every request | 807ms first, then **27ms** (negative cache) |
| scenarios listing | 0.6–**7.0s**, every request | 735ms first (`gh_list` now visible), then **27ms** |
| changed-laws | 0.6–1.0s per TTL window | 891ms first, then **29ms** |
| law body (warm) | 3ms | 3–9ms (unchanged) |
| save scenario → GET file | n/a | **28ms**, new content (read-your-writes) |
| save scenario → listing | n/a | in-place update, new file visible immediately |

No `refresh=` phase appeared synchronously in any probed request.

**One regression found and fixed** (pushed as a second commit): after a scenario DELETE, the invalidate-and-rebuild listing raced GitHub's eventually-consistent directory view — the rebuild still saw the deleted file and cached that stale listing for the rest of the snapshot window (`goneAfterDelete: false` in the probe). Fixed by updating the cached listing **in place** on save/delete (with a seed-probe on federated laws so a seed copy stays listed) and by skipping ghost entries (listed but content-read `None`) during a rebuild. Will re-verify the delete flow on the refreshed preview.

tdjager:
Re-verified on the refreshed pr929 preview (second commit): save → listing shows the file immediately (29ms, in-place update); **delete → listing drops the file immediately** (28ms — the GitHub directory-listing lag no longer leaks into the cache); GET on the deleted file 404s. Cold traject build unchanged (~5s, one-time per pod); warm listing/annotations stay ~27–32ms. This PR is ready for review.

tdjager:
Smoketested on the pr929 preview after bb04015f: scenario save→list→delete→list on a traject law (federated source) — file appears in the listing immediately after save, is gone immediately after delete (and stays gone; content GET 404s). Warm sidecar reads ~25-40ms. Marking ready for review.

PR #918, opened 2026-07-08
feat(editor): upload werkdocumenten with LLM conversion to markdown

tdjager:
## What

Users can add a document to a traject's **werkdocumenten** by **uploading** it. Supported in phase 1: **PDF and Word (.doc/.docx)**. The uploaded file is converted to markdown **asynchronously** and the result is stored as a normal `.md` werkdocument via the existing documents git backend. The werkdocumenten screen shows the status of running and failed conversions.

Known formats are converted **deterministically** with a real tool baked into the worker image — `pandoc` for Word/office formats, `pdftotext` (poppler) for PDF. Only when no tool fits does it fall back to the **agentic** path (the same `claude`/`opencode` subprocess the enrich pipeline uses, which decides for itself how to convert). Deterministic-first is reliable, needs no network egress, and keeps the untrusted document away from the Bash-enabled LLM — shrinking the prompt-injection surface (per the security note below).

## How

**Pipeline** (`packages/pipeline`)
- New `document_convert` job type + migrations: `jobs.traject_ref` (generic job↔traject link — the foundation per-traject harvest/enrich will also use) and a `document_uploads` (bytea) table. ZAD has no blob storage, so the binary rides transiently in Postgres and is deleted after conversion / terminal failure.
- Extracted `run_llm_subprocess` from `ProcessLlmRunner::run` so enrich and document-convert share the exact spawn/drain/timeout/memory-watchdog machinery (no enrich-test change).
- `document_convert` module: payload, upload storage helpers, status-list query, deterministic conversion (`pandoc`/`pdftotext`, baked into the enrich-worker image) with the agentic `LlmDocumentConverter` as fallback, and a service-account git write-back to the traject's writable-own branch (no PR — the async worker has no user session, mirroring how enrich commits).
- The enrich worker processes document-convert jobs **ahead of** the enrich hourly-cap gate, so an interactive upload is never blocked by the bulk-enrich budget.

**editor-api** (`packages/editor-api`)
- `POST /api/trajects/{ref}/corpus/documents/upload` — multipart PDF/Word upload; stores bytes, derives a collision-safe `.md` target and enqueues a `document_convert` job (`202 Accepted`). 25 MiB body cap.
- `GET /api/trajects/{ref}/corpus/documents/jobs` — running/failed conversion jobs for the status block.

**Frontend** (`frontend`)
[...]

PR #935, opened 2026-07-13
feat(editor): personal review tasks for user-requested async jobs

tdjager:
## Why

Users want to trigger async work from the editor (enrich a law; convert uploaded werkdocumenten) — but background workers have no user context. With per-user GitHub OAuth (#887) the user's token lives only in a sealed, request-scoped browser cookie, so **a background job can never commit as the user** (and document upload is even refused with 403 once user-token enforcement is on). On top of that, a terminally failed document conversion stays in the werkdocumenten list forever with no way to dismiss it.

Team discussion concluded: let the job run in the background, then hand the result back to the requester as a **personal task** to review and approve — bringing the commit back into the user's request context.

## What

**Task mechanism (generic) with user-requested enrich as the first consumer:**

- The worker never touches GitHub in the task flow. The editor-api snapshots the law YAML into Postgres (`job_blobs`, pattern: `document_uploads`) at request time; the enrichworker materialises it into a tempdir, enriches, and writes the result back as blobs — works for private traject repos where the worker has no token at all.
- On completion the worker creates a `tasks` row (migration `0028`) **atomically with `complete_job`**: personal, assigned to the requester, with an audit trail that survives resolution (only blobs are cleaned up; 7-day GC catches orphans).
- The editor shows a tasks button + badge (real `COUNT`) and a tasks sheet (NLDD). A `job_review` task opens the law in the editor with the proposal as an unsaved change; **Opslaan saves the full proposal via the normal law-save path** (operator token today, the user's OAuth token once `GITHUB_USER_TOKEN_REQUIRED` is on) and then resolves the task as approved. Verwerpen rejects it. Staleness (law changed since the snapshot) is detected via the existing `document_etag` and shown as a warning.
- Terminal failures become dismissable `job_failed` tasks — for enrich **and** for document conversion (`requested_by` is now recorded on upload). Failed conversions are gone from the werkdocumenten jobs list; they live in the tasks sheet instead.
- Robustness from the review rounds: retry classification mirrors the corpus path (deterministic content failures are terminal; resource exhaustion feeds the fork-exhaustion circuit breaker; transients retry, task jobs get `max_attempts=3` and re-materialise from blobs), `ensure_skills` runs on the task path, fail-marking + task creation are transactional, per-account cap of 20 active task jobs (429), `TASK_ENRICH_PROVIDER` validated at startup, and migration `0029` scopes the unique active-enrich index per traject so corpus-wide jobs and traject requests no longer block each other.

Everything is behind the (dual-registered) feature flag `tasks.job_review`, default off.

## ⚠️ Deploy note (migration 0029)

`0029` drops and recreates `idx_unique_active_enrich_job` with a fourth key expression. Binaries built from `main` use an `ON CONFLICT` target that only matches the **old** index — between the migration running and all Rust components being redeployed, admin bulk-enrich and the harvest→auto-enrich chain would error on enqueue. Deploy all Rust components together (the normal full deploy does this); avoid out-of-band single-component deploys while rolling this out. The index rebuild also takes a brief exclusive lock on `jobs` during the migration — subsecond at current table size, but it briefly blocks enqueues/claims.

## Custom CSS (project rule: must be reported)

- `TasksButton.vue`: wrapper (`position: relative`) + corner positioning for the count badge — `nldd-badge` has no built-in overlay placement relative to another element.
- `TasksSheet.vue`: `.tasks-list` flex column with `gap: 8px` (same stacking pattern as `ConversionStatus.vue`, previously reported).

## Out of scope / follow-ups

1. Document-convert **success** path via review tasks (would remove the per-traject `CORPUS_AUTH_*` push-token requirement entirely).
2. Traject-wide tasks (schema is ready: nullable assignee + separate `resolved_by`) and a diff view on task detail.
3. Progress relay on the enrich task path (the corpus path polls a progress file; the task path shows nothing while a human is waiting).
4. Reaper awareness: a task-flow job that dies without reaching a handler (worker crash, reap-exhausted) fails without a task; a sweep could create the missing `job_failed` task.
5. Edit-then-approve: manual edits on top of a proposal are not merged (the banner warns about this); merging would need a splice-onto-proposal step.
6. Task-list pagination beyond 100 open tasks (the badge count is already exact); account deletion with open tasks (NULL-assignee rows are unreachable in v1).
7. Editor-api handler tests for the new endpoints (blocked on editor-api tests reaching CI, see #931); store-layer coverage lives in `tasks_tests.rs` (9 integration tests).

## Testing

- `just check` green; pipeline integration tests (incl. 9 new `tasks_tests`) green; editor-api lib/bin + integration test targets green (except `save_annotations_test`, already broken on `main`, untouched — fixed by #931).
- Frontend: 498/498 vitest, build clean.
- Smoke test on the preview deployment follows in this PR (flag on → verrijk een wet → taak verschijnt → beoordelen → opslaan → commit; failed upload → dismissable task).

Design: spec + implementation plan in the team workspace (`superpowers-specs/2026-07-13-taken-mechanisme-*.md`); discussion refs: #887, #918.


tdjager:
## Preview smoke test (pr935) — results

Verified on the preview deployment with the shared test user:

- ✅ Feature flag `tasks.job_review`: PUT accepted (dual registration works, no silent revert); Taken button + empty sheet appear only with flag on + authenticated.
- ✅ "Verrijk deze wet" (pane menu) → "Verrijking gestart…" feedback; job enqueued with the exact designed payload (`deliver: task`, `requested_by`, `source_etag`, synthetic `yaml_path`, priority 80, `max_attempts` 3).
- ✅ Idempotency: second request for the same law → 409 with a clear message.
- ✅ Retry/re-materialisation: attempt 1 was reaped (`reaped: job stuck in processing`); attempts 2 and 3 automatically re-materialised the workdir from the input blob, as designed.
- ⚠️ **Full round-trip (task → review → approve) could not be completed live**: all three attempts were reaped after ~30 min each — the preview enrichworker currently never finishes an LLM run. This is a pre-existing environment issue (the job list shows several historical corpus enrich jobs with the same triple-reap pattern, e.g. BWBR0007795/BWBR0022751/BWBR0017017), likely the known worker OOM/hang failure mode — not a code path of this PR.
- 📌 This also empirically confirms follow-up #4 from the PR description: a triple-reaped task-flow job fails **without** a `job_failed` task (silent for the requester). That follow-up now has priority evidence.

The review/approve UI flow itself is covered by the unit/component test suite (498 tests incl. full-proposal save with If-Match, staleness banner, resolve ordering). Recommend re-running the round-trip once the preview worker is healthy, ideally with a small law.

tdjager:
## Smoke test update — full round-trip completed ✅

Follow-up to the earlier smoke-test comment: with a healthy enrichworker the complete round-trip has now been verified live on the preview, by the PR author's own account, on a **private-repo traject with user-token enforcement enabled**:

1. "Verrijk deze wet" on `besluit_zorgverzekering` → job completed in one attempt (354 articles, 8 newly machine-readable).
2. Review task appeared → "Beoordelen" opened the proposal in the editor → **Opslaan approved the full proposal, and the resulting commit is attributed to the requesting user** (their linked GitHub account) — the core design goal of this PR.

Also added since the last review round (user feedback from testing): a **"Bezig" section** in the tasks sheet showing the user's own running task-flow requests (spinner, no actions), plus a neutral dot on the tasks button when work is running but no action is needed — previously nothing was visible between requesting an enrichment and the task appearing.

PR #940, opened 2026-07-16
fix(frontend): home tab re-stamps the active traject

tdjager:
## What

The Home tab restored the last-visited Home path verbatim, so switching from a traject-scoped editor back to Home could land you in a stale scope - e.g. Corpus juris after opening a public deep-link - where the traject menu items (Instellingen / Werkdocumenten / Taken) don't exist.

The Home tab now re-stamps the **active** traject onto the stored path via a new `homeTabTarget` helper, mirroring what the Editor tab already does through `sectionTarget`:

- Editor in traject X → Home now lands in traject X's Home (same law preserved), not in whatever scope Home last had.
- When the stored path already carries the active scope it is still restored **verbatim**, so query/hash state (the open werkdocument's `?task=`, the `#yaml` detail tab) survives exactly like before.
- Werkdocumenten sub-mode keeps being preserved/re-stamped by the existing `sectionTarget` branches.

## Tests
- New `homeTabTarget` cases in `router.test.js` (verbatim same-scope incl. hash, re-stamp onto public path, re-stamp over stale traject, drop without traject).
- Full frontend vitest suite green (545 tests).

Follow-up from #939 (independent; no code dependency).

## Smoketest (pr940 preview)

Opened the public `/corpus-juris/wet_op_de_zorgtoeslag` deep-link (Home records it), switched to the editor in the "zorgtoeslag private repo" traject, then clicked the Home tab. It now lands on `/trajecten/{ref}/corpus/wet_op_de_zorgtoeslag` - traject scope with its sidebar menu, same law still open - instead of the traject-less Corpus juris page:

Home tab lands in the active traject

### DIFF
--- a/frontend/src/AppShell.vue
+++ b/frontend/src/AppShell.vue
@@ -14,6 +14,7 @@ import {
   lastHomePath,
   lastEditorPath,
   sectionTarget,
+  homeTabTarget,
   isHomeSection,
   rememberHarvesterOrigin,
 } from './composables/useLastVisitedRoute.js';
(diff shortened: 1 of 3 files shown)


tdjager:
Re: instellingen sub-mode dropped on traject mismatch — fixed in b4d01b36: sectionTarget now has an instellingen-traject branch mirroring the werkdocumenten one (re-stamp the active traject, keep the open tab), covered by new sectionTarget + homeTabTarget tests.

